PR #77, opened 2022-05-04
Run bundle add with group=development

vinistock:
A tiny improvement: when prompting for adding the `ruby-lsp` to the bundle, we weren't using `group=development` as an option.

We don't want the gem to be added to all groups, so we better scope it to development.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -123,7 +123,7 @@ export default class Client {
     );
 
     if (response === "Run bundle add and install") {
-      await this.execInPath("bundle add ruby-lsp");
+      await this.execInPath("bundle add ruby-lsp --group=development");
       await this.execInPath("bundle install");
       return false;
     }


KaanOzkan (approved):

vinistock:
I don't think there's a need for deleting the group information. We use it in [Tapioca](https://github.com/Shopify/tapioca/blob/main/Gemfile) and it seems to work fine.

st0012:
Not saying it'd break anything if not deleted. But I think it's relatively rare to see gems having the `development` group (at least not in [Rails](https://github.com/rails/rails/blob/main/Gemfile), [Sinatra](https://github.com/sinatra/sinatra/blob/master/Gemfile), [ruby/debug](https://github.com/ruby/debug/blob/master/Gemfile), nor in the gems [I maintain](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/Gemfile)). And in those cases, leaving the development group information will cause confusion, so I'd delete it.

PR #79, opened 2022-05-05
Prompt users to remove rubocop-lsp if present

vinistock:
Now that the `ruby-lsp` supports all features from the `rubocop-lsp`, we can help users cleanup their apps by prompting them to remove the gem.

This PR does two things
1. Inverts the order of the gem checks. Bundle list actually fails if gems aren't installed, so we should do `bundle check` first
2. Adds a prompt if `rubocop-lsp` is a part of the bundle

### Manual tests

1. Start the plugin on this branch
2. On the opened window, select a project that has the `rubocop-lsp` installed
3. Verify you get prompted

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -68,7 +68,7 @@ export default class Client {
   }
 
   async start() {
-    if ((await this.gemMissing()) || (await this.gemNotInstalled())) {
+    if ((await this.gemNotInstalled()) || (await this.gemMissing())) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -108,6 +108,19 @@ export default class Client {
   private async gemMissing(): Promise<boolean> {
     const bundledGems = await this.execInPath("bundle list");
 
+    if (bundledGems.includes("rubocop-lsp")) {
+      vscode.window.showErrorMessage(
+        "The rubocop-lsp gem has been replaced by the ruby-lsp.",
+        {
+          modal: true,
+          detail:
+            "Please, remove the rubocop-lsp and add ruby-lsp to your Gemfile",
```
egiurleo, line 117: This is an extreme nitpick but I think we should remove the comma after `Please`.

egiurleo (approved):

src/client.ts
vinistock: I changed it so that it displays the warning and then continues to check for the `ruby-lsp`. If it's not in the `Gemfile`, it'll still prompt people to add it to the bundle.

PR #85, opened 2022-05-16
Remove plugin order dependency from telemetry

vinistock:
### Motivation

Originally, we were trying to load the private telemetry API when activating the extension and then, if the command was present, telling the server that telemetry was enabled.

However, this approach is brittle because it makes the telemetry depend on the load order of the plugins. If a private metrics plugin is loaded after the Ruby LSP plugin, then the command will not have been registered yet and we fail to inform the server that there is a telemetry API.

### Implementation

I believe the most robust option is to have the server always push `telemetry/event` notifications (removing the concept of enabling the server telemetry).

Then here, on the plugin side, we try to initialize the metrics API object via the command when trying to send metrics. If we're unable to get an API object after 5 attempts, then it probably means there is no private metrics plugin and we can stop re-trying.

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -22,6 +22,7 @@ class DevelopmentApi implements TelemetryApi {
 
 export class Telemetry {
   private api?: TelemetryApi;
+  private initializationAttempts = 0;
 
   constructor(context: vscode.ExtensionContext, api?: TelemetryApi) {
     if (context.extensionMode === vscode.ExtensionMode.Development && !api) {
@@ -31,23 +32,29 @@ export class Telemetry {
     }
   }
 
-  async initialize() {
+  async sendEvent(event: TelemetryEvent) {
+    if (await this.initialize()) {
+      return this.api!.sendEvent(event);
+    }
+  }
+
+  private async initialize(): Promise<boolean> {
+    if (this.initializationAttempts > 5) {
+      return false;
+    }
+
     try {
-      this.api = await vscode.commands.executeCommand(
-        "ruby-lsp.getPrivateTelemetryApi"
-      );
+      if (!this.api) {
+        this.api = await vscode.commands.executeCommand(
+          "ruby-lsp.getPrivateTelemetryApi"
+        );
+        this.initializationAttempts++;
+      }
+
+      return Boolean(this.api);
     } catch (_error) {
       // Do nothing if no telemetry api is available
+      return false;
     }
-  }
-
-  async sendEvent(event: TelemetryEvent) {
-    if (this.api) {
-      return this.api.sendEvent(event);
-    }
-  }
-
-  enabled(): boolean {
-    return Boolean(this.api);
   }
 }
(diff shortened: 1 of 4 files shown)


src/telemetry.ts
```diff
@@ -29,23 +30,29 @@ export class Telemetry {
...
   }
 
-  async initialize() {
-    try {
-      this.api = await vscode.commands.executeCommand(
-        "ruby-lsp.getPrivateTelemetryApi"
-      );
-    } catch (_error) {
-      // Do nothing if no telemetry api is available
+  async sendEvent(event: TelemetryEvent) {
```
egiurleo, line 33: This approach makes sense to me overall, but I have a question - if we're moving the call to `initialize` inside the `sendEvent` method, why do we need to check for the telemetry API five times? Shouldn't all the APIs be loaded at this point if they exist? (Or maybe I'm misunderstanding the order of operations.)

vinistock: My idea was to not keep trying to execute the command forever, if someone is using the Ruby LSP without a private metrics plugin attached.

We know that the command will be loaded within 5 tries, so if we can't find the command after that, it's probably because the user doesn't have a metrics plugin - and not because it hasn't been loaded yet.

Morriar: Can't we check in `vscode.extensions.all` if the extension is there instead and keep trying if we found it? Or maybe this list would not be up-to-date if the extension is not loaded?

vinistock: That would remove the ability of other developers/companies adding their own private metrics plugin, as we wouldn't know the extension name or identifier.

Morriar: With the current approach, isn't there a change we're going to reach the limit before the other extension loads? What if I have 5 tabs opened already? Wouldn't I risk sending 5 `didOpen` events before the other extension loads?

Morriar: > Maybe we should just try to execute the command every time?

Would that be a noticeable overhead? Do we have a way to measure it?

vinistock: I don't think so. That's what Sorbet does and it doesn't seem to have a big impact. I'll switch the implementation.

src/telemetry.ts
```diff
@@ -31,23 +31,24 @@ export class Telemetry {
...
+  private async initialize(): Promise<boolean> {
+    try {
+      if (!this.api) {
+        this.api = await vscode.commands.executeCommand(
+          "ruby-lsp.getPrivateTelemetryApi"
+        );
+      }
+
+      return Boolean(this.api);
+    } catch (_error) {
```
Morriar, line 49: Should we be more precise regarding the of errors we swallow here?

vinistock: Is that possible in JavaScript?

Morriar: Kind of: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#conditional_catch-blocks

PR #151, opened 2022-07-11
Document version managers, commands and configurations

vinistock:
Improve our documentation about supported version managers, configurations and commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@ Search for `ruby-lsp` in the extensions tab and click install.
 
 **Note**
 
-For this extension to properly start the Ruby LSP server, the right Ruby version must be activated or else it can't find
-the right gems. Currently, auto activation is only supported for `shadowenv`, but support will be added for other Ruby
-version managers soon.
+For this extension to properly start the Ruby LSP server, the right Ruby version for the project being worked on must be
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,14 +4,49 @@
 
 ![Build Status](https://github.com/Shopify/vscode-ruby-lsp/workflows/CI/badge.svg)
 
-# vscode-ruby-lsp
```
st0012, line 7: I prefer to keep the old name. `Ruby LSP` is same as the actual `ruby-lsp`'s, which is confusing.

Or let's make it `Ruby LSP (VS Code extension)`?

README.md
```diff
@@ -4,14 +4,49 @@
...
 
 ## Usage
 
 Search for `ruby-lsp` in the extensions tab and click install.
 
+**Note**
+
+For this extension to properly start the Ruby LSP server, the right Ruby version must be activated or else it can't find
+the right gems. Currently, auto activation is only supported for `shadowenv`, but support will be added for other Ruby
+version managers soon.
```
adisonlampert, line 19: This might be a dumb q, but how do you figure out "the right Ruby version"?

adisonlampert (approved):

PR #167, opened 2022-07-25
Auto-restart the LSP on .rubocop.yml config changes

vinistock:
If the RuboCop configuration changes, then we have to restart the LSP to reload the diagnostics and formatting runners. Otherwise, we keep processing files using the old configuration.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -177,7 +177,7 @@ export default class Client {
 
   private registerAutoRestarts() {
     this.createRestartWatcher("Gemfile.lock");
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -177,6 +177,7 @@ export default class Client {
 
   private registerAutoRestarts() {
     this.createRestartWatcher("Gemfile.lock");
+    this.createRestartWatcher(".rubocop.yml");
```
Morriar, line 180: Should we also restart if a config file changed (or was added/removed) in a sub-directory?

PR #202, opened 2022-08-25
Add inlay hint support

RyanBrushett:
Necessary for the client to support inlay hints!


RyanBrushett:
See https://github.com/Shopify/ruby-lsp/issues/55

RyanBrushett:
Thanks for the review! Check it out @vinistock, this what you had in mind?

PR #226, opened 2022-09-15
Add support for chruby

vinistock:
First step towards #157

Add support for chruby. I'm proposing that we make the version manager and path to executable configurable, since trying to guess those can become quite messy.

For the chruby activation, we
1. Read the `.ruby-version` file to figure out which version to activate
2. Source the chruby script
3. Invoke the chruby script with the version we read
4. Echo all of the relevant environment variables. This is so that we can capture the results and insert them into the nodejs environment (otherwise all environment variable changes made by chruby are not applied to the parent process)

### Manual tests

1. On this branch, change your configuration to use chruby
```jsonc
"rubyLsp.rubyVersionManager": {
  "manager": "chruby",
  "path": "/opt/homebrew/opt/chruby/chruby.sh" // Find the actual path in your machine
}
```
2. Start the extension with F5 on the `ruby-lsp` repo
3. Verify you get an error prompt because no `.ruby-version` file exists
4. Create a `.ruby-version` file with some Ruby version (e.g.: `2.7.6`)
5. Reload the extension
6. Verify no errors pop up this time
7. In the output tab for the Ruby LSP, verify you get warnings about pattern matching being experimental (which are only printed in Ruby 2.7 and demonstrates that we selected the correct version)

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -1,0 +1,73 @@
+import { exec } from "child_process";
+import { promisify } from "util";
+import * as fs from "fs";
+
+import * as vscode from "vscode";
+
+const asyncExec = promisify(exec);
+const asyncReadFile = promisify(fs.readFile);
+
+export class Ruby {
+  private workingFolder: string;
+  private managerConfiguration: { [key: string]: string };
+
+  constructor() {
+    this.workingFolder = vscode.workspace.workspaceFolders![0].uri.fsPath;
+
+    this.managerConfiguration = vscode.workspace
+      .getConfiguration("rubyLsp")
+      .get("rubyVersionManager")!;
+  }
+
+  async activateRuby() {
+    switch (this.managerConfiguration.manager) {
+      case "chruby":
+        await this.activateChruby();
+        break;
+      default:
+        await this.activateShadowenv();
+        break;
+    }
+  }
+
+  async activateShadowenv() {
+    await vscode.extensions
+      .getExtension("shopify.vscode-shadowenv")
+      ?.activate();
+  }
+
+  async activateChruby() {
+    if (!fs.existsSync(`${this.workingFolder}/.ruby-version`)) {
+      vscode.window.showErrorMessage(
+        "Attempted to activate chruby environment, but no .ruby-version file was found."
+      );
+      return;
+    }
+
+    if (!fs.existsSync(this.managerConfiguration.path)) {
+      vscode.window.showErrorMessage(
+        `Attempted to activate chruby environment, but the path provided does not exist (${this.managerConfiguration.path}).`
+      );
+      return;
+    }
+
+    const rubyVersion = await asyncReadFile(
+      `${this.workingFolder}/.ruby-version`
+    );
+
+    try {
+      const result = await asyncExec(
+        `source ${this.managerConfiguration.path} &&
+         chruby "${rubyVersion.toString().trim()}" &&
+         ruby -rjson -e "puts JSON.dump(ENV.to_h)"`
+      );
+
+      // eslint-disable-next-line no-process-env
+      process.env = JSON.parse(result.stdout);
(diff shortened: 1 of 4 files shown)


src/ruby.ts
```diff
@@ -0,0 +1,83 @@
...
+        chruby "${rubyVersion.toString().trim()}" &&
+        echo "RUBY_ROOT=$RUBY_ROOT" &&
+        echo "RUBY_ENGINE=$RUBY_ENGINE" &&
+        echo "RUBYOPT=$RUBYOPT" &&
+        echo "PATH=$PATH" &&
+        echo "GEM_HOME=$GEM_HOME" &&
+        echo "GEM_ROOT=$GEM_ROOT" &&
+        echo "GEM_PATH=$GEM_PATH" &&
+        echo "RUBY_VERSION=$RUBY_VERSION"`
+      );
```
paracycle, line 69: Would it make more sense to somehow diff the environment before and after running `chruby` so that we can do all the same changes on `process.env`, instead of keeping a list of env variables we know will change?

I am thinking that we can do a `process.env` before calling this script, and we can do a `ruby -rjson -e "puts JSON.dump(ENV.to_h)"` inside the script (since Ruby should be properly setup now) and JSON parse the output of that in the JS script, so that we can compare the two envs and apply changes.

That could also be a reusable thing to apply around anything like this.

src/ruby.ts
```diff
@@ -0,0 +1,83 @@
...
+      return;
+    }
+
+    if (!fs.existsSync(this.managerConfiguration.path)) {
+      vscode.window.showErrorMessage(
+        `Attempted to activate chruby environment, but the path provided does not exist (${this.managerConfiguration.path}).`
+      );
+      return;
+    }
```
paracycle, line 50: Can we not make the `path` optional and expect to find it on the default path? I imagine that if someone is using `chruby` to manage their Ruby versions, they would have it properly setup and running `chruby <version>` would just work.

src/ruby.ts
```diff
@@ -0,0 +1,83 @@
...
+      .get("rubyVersionManager")!;
+  }
+
+  async activateRuby() {
+    switch (this.managerConfiguration.manager) {
+      case "chruby":
+        await this.activateChruby();
+        break;
+      default:
+        this.activateShadowenv();
```
paracycle, line 28: I think we want to `await` this method as well, since activate(): [Thenable](https://code.visualstudio.com/api/references/vscode-api#Thenable)<[T](https://code.visualstudio.com/api/references/vscode-api#T)>, `Extension<T>.activate` returns a promise (`Thenable<T>`) and if we return immediate, we would never be waiting for the activation to be finished.

I think that was our earlier problem and we needed to add a small delay to compensate for that.

vinistock: I've tried awaiting, but we still need the delay. There might be an `await` missing somewhere in vscode-shadowenv. I added the await anyways, but we'll keep the delay.

src/ruby.ts
vinistock: Chruby is installed in different paths on Linux and macos, so we'd need to look it up based on the architecture.

Also, I'm not sure the nodejs will inherit `chruby`, since it's not an executable. It's a bash function defined when you source the `chruby.sh` script. You think we don't need to source it and it should just work?

src/ruby.ts
paracycle: Hmm, it seems we don't even need to diff anything, we can just inherit the environment from the script:
```ts
const result = await asyncExec(`.....; ruby -rjson -e "puts JSON.dump(ENV.to_h)"`);
process.env = JSON.parse(result.stdout);
```

st0012 (approved):

paracycle (approved):
Amazing, thanks for finding a way to get this done with minimal config ❤️

PR #232, opened 2022-09-23
Automate release

vinistock:
Attempt to fully automate the release of our VS Code extensions.

This GitHub action can only be triggered manually and accepts a version as an input in the format `0.1.2`. It will then
- Use `sed` to change the version in `package.json`
- Commit the changes and push to main
- Create a tag based on main using the selected version
- Push the tags
- Create a GitHub release using the recently created tag

Creating the GitHub release will trigger the [publish action](https://github.com/Shopify/vscode-ruby-lsp/blob/main/.github/workflows/publish.yml), which then packages the extension and publishes in the marketplace.

If this works, then releasing our extensions becomes just a matter of triggering the workflow with the desired version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,46 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "The version to be released in the format X.Y.Z"
+        required: true
+        type: string
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
+name: Release
```
KaanOzkan, line 1: We already have a action named `Release` it could cause issues https://github.com/Shopify/vscode-ruby-lsp/blob/main/.github/workflows/publish.yml#L1

KaanOzkan (approved):

PR #251, opened 2022-10-12
Make gem update prompt quieter

egiurleo:
The Ruby LSP will prompts users to update the gem if it isn't already up-to-date with an intrusive pop-up. This has been distracting and confusing users. Instead, we can implement this functionality via an unobtrusive addition to the status bar. When the user clicks the button in the status bar, it will update the gem for them automatically and then disappear.

Screen Recording 2022-10-12 at 1 56 15 PM


src/client.ts
```diff
@@ -144,6 +142,20 @@ export default class Client {
     await this.start();
   }
 
+  async updateGem(): Promise<void> {
```
egiurleo, line 145: One question I had while making this PR is how much of the functionality to include in the client vs in the `extension.ts` file. In the end, I chose to implement the state in the client and the display in `extension.ts`. Happy to change it and put all the functionality back in the client if that works better!

src/extension.ts
```diff
@@ -28,3 +31,36 @@ export async function deactivate(): Promise<void> {
...
+
+async function activateGemOutdatedButton(context: vscode.ExtensionContext) {
+  const commandId = "updateOutdatedGem";
+
+  gemOutdatedButton = vscode.window.createStatusBarItem(
+    vscode.StatusBarAlignment.Left,
+    0
+  );
+
+  gemOutdatedButton.text = "$(plug) Update Ruby LSP";
```
egiurleo, line 43: I chose a little plug icon because it looks like this extension's logo! Let me know what you think.

Screen Shot 2022-10-12 at 2 03 01 PM

src/extension.ts
```diff
@@ -28,3 +31,36 @@ export async function deactivate(): Promise<void> {
...
+  gemOutdatedButton.tooltip =
+    "The Ruby LSP gem is not up-to-date. Press this button to install the latest version of the gem.";
+  gemOutdatedButton.command = commandId;
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand(commandId, async () => {
+      await client.updateGem();
+      await updateGemOutdatedButton();
```
egiurleo, line 51: If the gem is successfully updated, this will hide the button from the status bar so it's no longer bothering the user.

src/bundler.ts
```diff
@@ -0,0 +1,28 @@
+import { exec } from "child_process";
```
egiurleo, line 1: @vinistock Is this what you were envisioning when you mentioned a `bundler.ts` file?

egiurleo: (And assuming you like this pattern I can make a follow up PR moving all the other bundler methods in here)

src/extension.ts
```diff
@@ -28,3 +31,38 @@ export async function deactivate(): Promise<void> {
...
+  gemOutdatedButton.command = commandId;
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand(commandId, async () => {
+      await updateGem();
+      await updateGemOutdatedButton(gemOutdatedButton);
+    })
+  );
+
+  updateGemOutdatedButton(gemOutdatedButton);
```
egiurleo, line 55: @vinistock I know you mentioned inlining these methods but I really didn't want to write out the exact same code twice, so I'm passing the button to the method instead of having it be a global variable. Let me know if this works okay for you.

src/bundler.ts
```diff
@@ -0,0 +1,28 @@
...
+    await execInPath("bundle outdated ruby-lsp");
+  } catch {
+    return true;
+  }
+
+  return false;
+}
+
+export async function updateGem(): Promise<void> {
+  await execInPath("bundle update ruby-lsp --conservative");
```
vinistock, line 19: I know this is not related to this PR, but can you please remove the conservative flag? I realized that if we updated the syntax_tree requirement, using conservative actually fails. Now that we're no longer putting the update window in people's faces, I think it's safe to use a regular update.

src/extension.ts
```diff
@@ -28,3 +31,38 @@ export async function deactivate(): Promise<void> {
...
+
+  gemOutdatedButton.text = "$(plug) Update Ruby LSP";
+  gemOutdatedButton.tooltip =
+    "The Ruby LSP gem is not up-to-date. Press this button to install the latest version of the gem.";
+  gemOutdatedButton.command = commandId;
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand(commandId, async () => {
+      await updateGem();
+      await updateGemOutdatedButton(gemOutdatedButton);
```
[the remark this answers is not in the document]
vinistock: We can use the same technique used [here](https://github.com/Shopify/vscode-ruby-lsp/pull/253), with a progress indicator.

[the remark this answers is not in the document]
egiurleo: That's so cool! I added the progress indicator. You can see it in the GIF in the README. If the gem updates successfully, the button will disappear from the status bar. If it doesn't (there's an error or the gem is pinned on an old version) then the button will remain.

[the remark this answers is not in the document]
vinistock: We could inspect the output of running bundle install and if it contains the message `Tried to update ruby-lsp, but ...` then we use the progress indicator to inform them of it. Something like

```typescript
(progress) => {
  if (output.includes("blah blah")) {
    progress.report({ message: "Could not update the ruby-lsp gem. Is the version in the Gemfile pinned?" })
  }
}
```

vinistock (approved):

rafaelfranca (approved):

src/extension.ts
```diff
@@ -28,3 +31,50 @@ export async function deactivate(): Promise<void> {
...
+    "The ruby-lsp gem is not up-to-date. Press this button to install the latest version of the gem.";
+  gemOutdatedButton.command = commandId;
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand(commandId, async () => {
+      gemOutdatedButton.text = "$(loading~spin) Updating Ruby LSP";
+      const result = await updateGem();
+
+      if (result.stderr.length > 0) {
+        gemOutdatedButton.text = "$(close) Ruby LSP Update Failed";
```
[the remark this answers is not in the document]
egiurleo: Actually no this feels correct to me. The three states of this button are:
- `Update Ruby LSP`
- `Updating Ruby LSP`
- `Ruby LSP Update Failed`

Maybe `Failed to update Ruby LSP`? What do you think?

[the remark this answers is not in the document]
Morriar: It just caught my eye. If it's intended, let's merge 👍

PR #254, opened 2022-10-14
Allow running the Ruby LSP server with YJIT

vinistock:
Starting on Ruby 3.2, YJIT provides a good performance boost on the Ruby LSP. This PR adds a new configuration to use YJIT and we should default to using it as long as the Ruby version is greater or equals to 3.2 and the Ruby was compiled with YJIT support.

The following chart is a comparison of the average time to execute requests with and without YJIT.

Ruby LSP + YJIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,11 @@
             "chruby"
           ],
           "default": "shadowenv"
+        },
+        "rubyLsp.yjit": {
+          "description": "Use YJIT to speed up the Ruby LSP server",
+          "type": "boolean",
+          "default": true
         }
       }
     },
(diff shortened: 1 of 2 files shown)


vinistock:
@Morriar done!

Morriar:
Results seem very encouraging!

egiurleo (approved):
🎉

st0012 (approved):

st0012:
Just curious: would enabling YJIT on Spin cause any issue due to the potential increase on memory consumption? I think the answer is **NO** but it's made based my very limited understandings:

1. YJIT may consume a bit more memory
2. But Spin should have more than enough memory for that ([reference](https://gist.github.com/campbel/87e86713803f0b482985999d4aa913a3))

So I'm curious if they're true or not.

vinistock:
Both assumptions are true. The Ruby LSP doesn't use a LOT of memory, so the increase resulting from YJIT should be okay.

PR #274, opened 2022-10-28
Always use ruby -v to find out the Ruby version

vinistock:
We were using the environment variable RUBY_VERSION in some cases, but that seems to not be set for every version manager. Specifically, ASDF does not set it, which ends up putting the extension in a weird state.


rzane:
Running `ruby --version` outputs a lot of information that probably isn't relevant. Here's how I obtained the Ruby version: https://github.com/Shopify/vscode-ruby-lsp/blob/e2100b57b092a8c26374150157b29cad10cac648/src/ruby.ts#L88-L91

vinistock:
Is that variable always set by Ruby? Or is it manager dependant?

src/ruby.ts
```diff
@@ -86,8 +87,13 @@ export class Ruby {
   }
 
   private async displayRubyVersion() {
-    const rubyVersion = await asyncExec('ruby -e "puts RUBY_VERSION"');
```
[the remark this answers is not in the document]
vinistock: Okay, so the `RUBY_VERSION` object is always set. We were using the environment variable with the same name to determine whether to start with YJIT or not, which was causing issues.

rzane:
Yes, it's a constant documented here: https://ruby-doc.org/core-3.1.2/Object.html. It shouldn't be dependent on the version manager at all. I think if it didn't work with ASDF, there must be something else was the real issue.

src/client.ts
```diff
@@ -282,7 +289,7 @@ export default class Client {
     const useYjit = vscode.workspace.getConfiguration("rubyLsp").get("yjit");
 
     // Enabling YJIT only provides a performance benefit on Ruby 3.2.0 and above
-    if (!useYjit || env.RUBY_VERSION! < "3.2.0") {
+    if (!useYjit || !this.ruby.rubyVersion || this.ruby.rubyVersion < "3.2.0") {
```
paracycle, line 292: I think we should use proper YJIT feature detection and not do Ruby version comparisons here. Something like `this.ruby.yjit_enabled` would be much nicer. It is as simple as checking if [`RubyVM::YJIT`](https://ruby-doc.org/core-3.1.2/RubyVM/YJIT.html) is defined using `defined?(RubyVM::YJIT)`

vinistock: Didn't know about the `RubyVM::YJIT` constant, thanks for the tip. For the version checking, I tried avoiding a new dependency and just compared the major and minor versions. Let me know what you think!

src/client.ts
```diff
@@ -281,14 +293,14 @@ export default class Client {
...
       return env;
     }
 
-    // Check if the Ruby was compiled with YJIT since it's not required. The exec must be sync here because this is used
-    // in the constructor, which cannot be an async function
-    if (!execSync("ruby -v --yjit").toString().includes("+YJIT")) {
+    const [major, minor, _patch] = this.ruby.rubyVersion.split(".");
+
+    // Enabling YJIT only provides a performance benefit on Ruby 3.2.0 and above
+    if (!useYjit || !this.ruby.yjitEnabled || !(major >= "3" && minor >= "2")) {
```
paracycle, line 303: I am afraid this is still not adequate since we are not comparing integers. `"10" > "2"` will still be `false`, for the minor comparison, for example.
```suggestion
    const [major, minor, _patch] = this.ruby.rubyVersion.split(".").map(Number);

    // Enabling YJIT only provides a performance benefit on Ruby 3.2.0 and above
    if (!useYjit || !this.ruby.yjitEnabled || (major == 3 && minor < 2)) {
```

paracycle: Or we can check it like `[major, minor] < [3, 2]` if you want.

paracycle:
```js
> [2, 1] < [3, 2]
true
> [3, 1] < [3, 2]
true
> [3, 2] < [3, 2]
false
> [4, 1] < [3, 2]
false
```

PR #309, opened 2022-12-06
Avoid shelling out multiple times when activating Ruby

rzane:
We can load all of the environment information we'll need in a single invocation of the ruby executable.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -1,4 +1,4 @@
-import { exec } from "child_process";
+import { exec, ExecOptions } from "child_process";
 import { promisify } from "util";
 import { readFile } from "fs/promises";
 
@@ -40,9 +40,7 @@ export class Ruby {
           break;
       }
 
-      await this.delay(500);
-      await this.displayRubyVersion();
-      this.checkYjit();
+      vscode.window.setStatusBarMessage(`Ruby ${this.rubyVersion}`);
     } catch (error: any) {
       vscode.window.showErrorMessage(
         `Failed to activate ${this.versionManager} environment: ${error.message}`
@@ -51,9 +49,13 @@ export class Ruby {
   }
 
   private async activateShadowenv() {
-    await vscode.extensions
-      .getExtension("shopify.vscode-shadowenv")
-      ?.activate();
+    const shadowenv = vscode.extensions.getExtension(
+      "shopify.vscode-shadowenv"
+    );
+
+    await shadowenv?.activate();
+    await this.delay(500);
+    await this.analyze("ruby");
   }
 
   private async activateChruby() {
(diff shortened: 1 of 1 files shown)


vinistock:
While I agree with the idea to avoid invoking the Ruby executable excessively, this breaks up a bit for shadowenv, since it has a different flow than the other version managers. The shadowenv extension updates the process environment, but we never update the Ruby version and YJIT attributes.

Maybe we can unify those two only?

src/ruby.ts
```diff
@@ -36,24 +36,29 @@ export class Ruby {
...
-      .getExtension("shopify.vscode-shadowenv")
-      ?.activate();
+  private async activateFallback() {
+    const shadowenv = vscode.extensions.getExtension(
+      "shopify.vscode-shadowenv"
+    );
+
+    if (shadowenv) {
+      await shadowenv.activate();
+      await this.delay(500);
```
rzane, line 58: This delay should only be necessary for shadowenv.

src/ruby.ts
```diff
@@ -36,24 +36,29 @@ export class Ruby {
...
         `Failed to activate ${this.versionManager} environment: ${error.message}`
       );
     }
   }
 
-  private async activateShadowenv() {
-    await vscode.extensions
-      .getExtension("shopify.vscode-shadowenv")
-      ?.activate();
+  private async activateFallback() {
```
vinistock, line 51: Let's continue to call it `activateShadowenv`.

src/ruby.ts
```diff
@@ -80,19 +85,28 @@ export class Ruby {
...
-    const rubyVersion = await asyncExec('ruby -e "puts RUBY_VERSION"');
-    this.rubyVersion = rubyVersion.stdout.trim();
-    vscode.window.setStatusBarMessage(`Ruby ${this.rubyVersion}`);
+  private async analyze(
+    ruby: string,
+    opts?: ExecOptions
+  ): Promise<{ [key: string]: string }> {
+    const result = await asyncExec(
+      `${ruby} -rjson -e "puts JSON.dump(env: ENV.to_h, version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT))"`,
+      { ...opts, encoding: "utf-8" }
```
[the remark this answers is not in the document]
rzane: I added this to appease typescript. If I don't specify the encoding, `result.stdout` is `string | Buffer`. If I specify utf-8, `result.stdout` is `string`.

src/ruby.ts
vinistock: Another option is to add another branch for shadowenv and have the default do nothing. In that case, we won't be using a version manager and will just pick up whatever is currently set in the VS Code environment.

vinistock:
Thank you for the contribution, but I think main diverged considerably since then.

Issue #262, opened 2022-10-19
Create a status bar button for centralized Ruby LSP interaction


vinistock:
We initially attempted to tell users about best practices and how to set up the Ruby LSP through `showInformation` dialogues. However, this does not seem to achieve the goal as we expected and we actually end up causing notification fatigue on developers.

It would be nice to unify all interaction with the Ruby LSP through a status bar button.

### Idea

The status bar item could
- Display the current status of the Ruby LSP server (ready, stopped, errored or not installed)
- Clicking on it would then open a `showInformation` dialogue with action options depending on the state
    - if the server is running, then show stop or restart
    - if the server gem is not in the bundle, then show bundle add
    - if the server has errored, then show a reload window button
    - if the server is stopped, then show a start button

### Note

The only interaction I believe should continue to be a notification is if the gems are not installed (bundle install prompt).

adisonlampert:
In addition to the features above, the status button will now include:
- The Ruby version
- Whether it’s using YJIT or not
- If the server gem is outdated

We are now implementing with the VS Code recommended [language status item](https://code.visualstudio.com/api/references/vscode-api#LanguageStatusItem).

joeyschoblaska:
Will these syntax error notifications be going away as well? I timed it and they stick around for _20 seconds_, whether you fix the error or not.

<img width="485" alt="image" src="https://user-images.githubusercontent.com/29129/215924943-6df5230a-a07f-46a8-88ad-ae0e111f232d.png">

vinistock:
The centralized place for interactions would be for displaying current information (e.g.: server is stopped, server has errored) and buttons to perform actions on it.

That specific warning is an error coming trying to execute the formatting request. There's no way to automatically dismiss it from the server or change how long they stick around, as far as I know.

Are you using the Ruby LSP with RuboCop? If so, you should see diagnostics ahead of time so that you can fix the syntax before trying to save the file.

PR #320, opened 2022-12-14
Create status items

adisonlampert:
Related to #262 

### Features 
This PR creates language status items to:
- Warn users if the server gem has not been installed
  - If it hasn't been installed, there's an option to run `bundle install`
- Warn users if `ruby-lsp` is not on the bundle list
  - If it hasn't been added, there's an option to run `bundle add`
- Warn users if `ruby-lsp` is out-of-date and prompt them to update
- If the server is running, allow users to stop it
- If the server is stopped, allow users to start it
- If the server has errored, allow users to restart it
  - This feature is not totally complete because I think it will require some refactoring in `client.ts` and the PR is already pretty big so I felt this would be best in a separate PR
 - Show what Ruby version is being used
 - Show if YJIT is being used with the LSP
   - If the user is using Ruby  >= 3.2 and has YJIT set up, they have the option to toggle YJIT.

### Appearance
To view the language status items, you must have a Ruby file open and you must hover over the `{ }` next to "Ruby"

Screenshot 2023-01-11 at 1 29 54 PM
Screenshot 2023-01-11 at 2 35 51 PM
Screenshot 2022-12-14 at 8 29 34 AM
Screenshot 2023-01-11 at 2 41 15 PM

**Note:** You can pin language status items and they will appear in your status bar. Once you pin it, the pin will remain and even if the language status item is disposed of, it'll be pinned if it is created again. However, the API does not currently allow developers to pin items automatically so this must be done by the user. Clicking the pinned language status item will trigger its command. It's nice because developers can customize what they'd like so their status bar doesn't get too noisy.

### Test
1. Open this branch in VSCode
2. Run the branch using the "Debug and Run" tab on the Workbench
3. Open a Ruby file
4. Check out the language status items by hovering over the `{ }` next to Ruby (see images above)
5. Try out starting and stopping the server (You can see that this is working by looking at the "Debug Console" in your `vscode-ruby-lsp` window)
6. Create some errors
    - Run `bundle remove ruby-lsp` in your repository that you're using to debug and then use the language status item to run `bundle add`
    - Run `gem uninstall ruby-lsp`, reload the debugging window, use the language status item to run `bundle install`
    - Install an older version of `ruby-lsp` and uninstall the most recent version, reload, and use the language status item to install the latest version (You may get a "The gems in this bundle are not installed" error if the version of `ruby-lsp` is specified in the `Gemfile` to be `~> 0.3.7`)
7. Toggle YJIT
    - Ensure you have YJIT
    - Open a project using Ruby `>= 3.2.0` (e.g., `ruby-lsp`)
    - You should see the option from the first screenshot. You can toggle from the status item.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -6,10 +6,12 @@ import * as vscode from "vscode";
 const asyncExec = promisify(exec);
 
 export async function isGemMissing(): Promise<boolean> {
-  const bundledGems = await execInPath("bundle list");
-  const hasRubyLsp = bundledGems.stdout.includes("ruby-lsp");
-
-  return !hasRubyLsp;
+  try {
+    await execInPath("bundle show ruby-lsp");
+    return false;
+  } catch {
+    return true;
+  }
 }
 
 export async function isGemOutdated(): Promise<boolean> {
@@ -23,8 +25,9 @@ export async function isGemOutdated(): Promise<boolean> {
 }
 
 export async function addGem(): Promise<void> {
-  await execInPath("bundle add ruby-lsp --group=development");
-  await execInPath("bundle install");
+  await execInPath(
+    "bundle add ruby-lsp --group=development --require=false && bundle install"
+  );
 }
 
 export async function updateGem(): Promise<{ stdout: string; stderr: string }> {
@@ -32,19 +35,12 @@ export async function updateGem(): Promise<{ stdout: string; stderr: string }> {
 }
 
 export async function bundleCheck(): Promise<boolean> {
-  let bundlerCheck: string;
-
   try {
-    bundlerCheck = (await execInPath("bundle check")).stdout;
+    await execInPath("bundle check");
+    return true;
   } catch {
-    bundlerCheck = "";
-  }
-
-  if (bundlerCheck.includes("The Gemfile's dependencies are satisfied")) {
     return false;
   }
-
-  return true;
 }
 
 export async function bundleInstall(): Promise<{
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -5,6 +5,13 @@ import * as vscode from "vscode";
 
 const asyncExec = promisify(exec);
 
+export async function isGemMissing(): Promise<boolean> {
+  const bundledGems = await execInPath("bundle list");
+  const hasRubyLsp = bundledGems.stdout.includes("ruby-lsp");
+
+  return !hasRubyLsp;
+}
+
 export async function isGemOutdated(): Promise<boolean> {
   try {
     await execInPath("bundle outdated ruby-lsp");
@@ -15,8 +22,36 @@ export async function isGemOutdated(): Promise<boolean> {
   return false;
 }
 
+export async function addGem(): Promise<void> {
+  await execInPath("bundle add ruby-lsp --group=development");
+  await execInPath("bundle install");
+}
+
 export async function updateGem(): Promise<{ stdout: string; stderr: string }> {
   return execInPath("bundle update ruby-lsp");
+}
+
+export async function bundleCheck(): Promise<boolean> {
+  let bundlerCheck: string;
+
+  try {
+    bundlerCheck = (await execInPath("bundle check")).stdout;
+  } catch {
+    bundlerCheck = "";
+  }
+
+  if (bundlerCheck.includes("The Gemfile's dependencies are satisfied")) {
+    return false;
+  }
+
+  return true;
+}
+
+export async function bundleInstall(): Promise<{
+  stdout: string;
+  stderr: string;
+}> {
+  return execInPath("bundle install");
 }
 
 async function execInPath(
--- a/src/status.ts
+++ b/src/status.ts
@@ -10,10 +10,10 @@ import {
 } from "./bundler";
 import { Ruby } from "./ruby";
 
-export const enum ServerCommand {
+export enum ServerCommand {
   Start = "ruby-lsp.start",
   Stop = "ruby-lsp.stop",
-  Error = "ruby-lsp.restart",
+  Restart = "ruby-lsp.restart",
 }
 
 export class StatusItem {
@@ -21,6 +21,7 @@ export class StatusItem {
   private ruby: Ruby;
   private selector: vscode.DocumentSelector;
   private serverStatus: vscode.LanguageStatusItem;
+  private yjitStatus: vscode.LanguageStatusItem;
 
   constructor(context: vscode.ExtensionContext, ruby: Ruby) {
     this.context = context;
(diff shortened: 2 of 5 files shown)


src/status.ts
```diff
@@ -0,0 +1,267 @@
...
+      }
+    }
+  }
+
+  private createRubyStatus(ruby: Ruby) {
+    const rubyVersion: vscode.LanguageStatusItem =
+      vscode.languages.createLanguageStatusItem("rubyVersion", this.selector);
+    rubyVersion.name = "Ruby LSP Status";
+    rubyVersion.text = `Using Ruby ${ruby.rubyVersion}`;
+  }
```
adisonlampert, line 253: I included this because it's part of the issue, but the Ruby version is already in the status bar (at least for me). Does anyone else already have the Ruby version in their status bar (left side, all the way to the right)? I don't know where this is coming from because I checked `vscode-ruby-shopify` and `sorbet-vscode` and from quickly glancing over that, I couldn't find this in the source code. If this is something most people already have visible, is this still necessary?

src/bundler.ts
```diff
@@ -15,10 +22,38 @@ export async function isGemOutdated(): Promise<boolean> {
...
+export async function bundleCheck(): Promise<boolean> {
+  let bundlerCheck: string;
+
+  try {
+    bundlerCheck = (await execInPath("bundle check")).stdout;
+  } catch {
+    bundlerCheck = "";
+  }
+
+  if (bundlerCheck.includes("The Gemfile's dependencies are satisfied")) {
```
andyw8, line 43: Is it possible to use the exit code rather than checking the text?

src/bundler.ts
```diff
@@ -15,10 +22,38 @@ export async function isGemOutdated(): Promise<boolean> {
   return false;
 }
 
+export async function addGem(): Promise<void> {
+  await execInPath("bundle add ruby-lsp --group=development");
+  await execInPath("bundle install");
```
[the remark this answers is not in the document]
vinistock: Also, in the case it does not, let's reuse the new bundleInstall function.

src/status.ts
```diff
@@ -0,0 +1,267 @@
...
+    if (this.serverStatus) {
+      this.serverStatus.dispose();
+    }
+
+    switch (status) {
+      case ServerCommand.Start: {
+        this.serverStatus = this.createStatusItem(
+          "serverStatus",
+          "Ruby LSP: Running...",
+          vscode.LanguageStatusSeverity?.Warning
```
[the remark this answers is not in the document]
vinistock: Does the order in which we create the language status items affect the order? Maybe we can just create them in the order we want them displayed.

src/status.ts
```diff
@@ -0,0 +1,267 @@
...
+
+        this.serverStatus.command = {
+          title: "Restart Server",
+          command: ServerCommand.Error,
+        };
+      }
+    }
+  }
+
+  public async installGems(): Promise<boolean> {
```
vinistock, line 99: Question about the UX for everybody: before we had the "gems are not installed" dialogue, people use to complain regularly that Sorbet or the Ruby LSP weren't working because they forgot to bundle install - which we haven't heard anymore since the dialogue is in place.

Do we risk falling back into that scenario by putting this in the language status item? Should we treat bundle install as the only special case in which we display a dialogue and put everything else in status items?

src/status.ts
```diff
@@ -0,0 +1,267 @@
...
+      const status: vscode.LanguageStatusItem = this.createStatusItem(
+        "installGems",
+        "Ruby LSP: The gems in the bundle are not installed.",
+        vscode.LanguageStatusSeverity?.Error
+      );
+
+      this.context.subscriptions.push(
+        vscode.commands.registerCommand("ruby-lsp.installGems", () => {
+          status.text = "Ruby LSP: Installing gems...";
+          status.command = undefined;
```
vinistock, line 111: I don't think you need to explicitly set it to `undefined`, that should be the default.

src/status.ts
```diff
@@ -0,0 +1,267 @@
...
+      "updateGem",
+      "Ruby LSP: The gem is not up-to-date",
+      vscode.LanguageStatusSeverity?.Warning
+    );
+
+    this.context.subscriptions.push(
+      vscode.commands.registerCommand(commandId, async () => {
+        status.text = "Ruby LSP: Updating Ruby LSP";
+        status.busy = true;
+        status.command = undefined;
```
vinistock, line 185: Same here.

src/status.ts
```diff
@@ -0,0 +1,267 @@
...
+
+        const result = await updateGem();
+
+        if (result.stderr.length > 0) {
+          status.text = "Ruby LSP: Update failed";
+          status.command = {
+            title: "Try again",
```
vinistock, line 192: Is there a scenario in which trying again can succeed? I feel like if bundle update fails, the dev is probably going to have to jump to the terminal and figure it out.

src/status.ts
```diff
@@ -0,0 +1,267 @@
...
+
+    if (this.ruby.rubyVersion) {
+      [major, minor] = this.ruby.rubyVersion.split(".").map(Number);
+    }
+
+    const yjit: vscode.LanguageStatusItem =
+      vscode.languages.createLanguageStatusItem("yjit", this.selector);
+    yjit.name = "Ruby LSP Status";
+
+    if (useYjit && ruby.yjitEnabled && [major, minor] >= [3, 2]) {
```
vinistock, line 232: Instead of duplicating this check, what do you think about setting some attribute on `Client` like `yjitEnabled`? Then instead of passing the client's context and Ruby to `Status`, we can pass the entire client object and make whatever we need to inspect public.

src/status.ts
```diff
@@ -0,0 +1,267 @@
...
+      yjit.text = "Ruby LSP: YJIT is in use";
+    } else {
+      yjit.text = "Ruby LSP: YJIT is not in use";
+      if ([major, minor] >= [3, 2]) {
+        yjit.severity = vscode.LanguageStatusSeverity?.Warning;
+        yjit.command = {
+          title: "Enable it",
+          command: "",
+        };
```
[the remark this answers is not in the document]
vinistock: That would depend on why YJIT is disabled. If it's just the configuration, then it's easy to offer to enabled that. However, if their Ruby was compiled without YJIT because they don't have the Rust compiler in their machine, then it might be trickier and, honestly, territory we don't want to venture into.

How about displaying enable it only if the configuration is set to false when YJIT is available and Ruby is > 3.2?

vinistock (commented):
This is looking great. Since we'll be displaying Ruby version in the status items, don't forget to remove the status bar item pushed [here](https://github.com/Shopify/vscode-ruby-lsp/blob/2c6c63970ca33ab4f395b72e10cffb64e0bcb3cd/src/ruby.ts#L95).

src/status.ts
egiurleo: I don't think I'm 100% on this opinion, but I'm leaning towards wanting it in the status item. Thinking about the Sorbet LSP, I like that it doesn't spam me, but I know where to look for feedback if Sorbet is not working. Plus the status being displayed in the status item makes it easy to see if something weird is going on.

src/status.ts
egiurleo: I also have the Ruby version in my status bar.

adisonlampert: LOL turns out it's in the one place I didn't check: this project. It's in `ruby.ts`. I'm removing it from there so now the Ruby version is only in one place in the status bar as a status item

src/status.ts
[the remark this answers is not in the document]
adisonlampert: So I think making it just `Information` is best so people can pin it if they want and it's not the ugly yellow color. However, it doesn't  look like the order of status items with the same severity is affected by the order it is added:
Screenshot 2023-01-11 at 1 22 31 PM
The server status should be the first item, but it's the second. It's not ordered alphabetically either as far as I can tell so I'm not sure why it's ordered this way, but I think this order is good?

src/status.ts
adisonlampert: Since that would require a bit of refactoring and this PR is already quite large, would it be best to put this into a separate issue or should it still be addressed here?

src/bundler.ts
paracycle: Just need to add `error: Error` to the object type in the `Promise` that is being returned, and check if `error` is `null` or not.

So: `: Promise<{ stdout: string; stderr: string }>` becomes `: Promise<{ stdout: string; stderr: string, error: Error }>` and `bundlerCheck` becomes:

```js
let bundlerCheck = (await execInPath("bundle check").error
```
and this line becomes:
```suggestion
  return !bundlerCheck;
```

paracycle: Ooops, I am sorry, what I wrote above isn't exactly correct, since I assumed too much about how `util.promisify` works.

According to the [docs](https://nodejs.org/docs/latest-v12.x/api/child_process.html#child_process_child_process_exec_command_options_callback):
> If this method is invoked as its [util.promisify()](https://nodejs.org/docs/latest-v12.x/api/util.html#util_util_promisify_original)ed version, it returns a `Promise` for an `Object` with `stdout` and `stderr` properties. The returned `ChildProcess` instance is attached to the `Promise` as a `child` property. In case of an error (including any error resulting in an exit code other than `0`), a rejected promise is returned, with the same `error` object given in the callback, but with two additional properties `stdout` and `stderr`.

So there is no need to change the return type of `execInPath` (and it would be wrong to change it anyway), the fact that an exception is raised from `await` and that we are catching it means that the exec failed. That should be good enough to decide if `bundle check` succeeded or not.

src/bundler.ts
```diff
@@ -15,10 +22,38 @@ export async function isGemOutdated(): Promise<boolean> {
...
+    bundlerCheck = (await execInPath("bundle check")).stdout;
+  } catch {
+    bundlerCheck = "";
+  }
+
+  if (bundlerCheck.includes("The Gemfile's dependencies are satisfied")) {
+    return false;
+  }
+
+  return true;
```
paracycle, line 47:
```suggestion
  try {
    await execInPath("bundle check");
    return true;
  } catch (e) {
    return e.code == 0;
  }
```

src/status.ts
```diff
@@ -0,0 +1,281 @@
+import * as vscode from "vscode";
+
+import {
+  isGemMissing,
+  addGem,
+  bundleCheck,
+  bundleInstall,
+  isGemOutdated,
+  updateGem,
+} from "./bundler";
```
paracycle, line 10: I would find:
```suggestion
import Bundler from "./bundler";
```
and then referring to methods like `Bundler.addGem` to be much more readable then importing loose functions.

paracycle (commented):
This looks really cool, thanks!

src/status.ts
adisonlampert: After discussing this in our team weekly sync, we decided it would be best to have the "gems are not installed" dialogue as a language status item for this PR and then in a future PR, especially if people begin having issues, investigate whether it would be better as a status bar item or revert it back to an information message.

src/bundler.ts
```diff
@@ -5,6 +5,13 @@ import * as vscode from "vscode";
 
 const asyncExec = promisify(exec);
 
+export async function isGemMissing(): Promise<boolean> {
+  const bundledGems = await execInPath("bundle list");
```
st0012, line 9: It looks like we used the `try catch` pattern in `isGemOutdated` and `bundleCheck` functions. Do we want to align with them here? For example:

```js
  try {
    await execInPath("bundle show ruby-lsp");
    return false;
  } catch (exc) {
    return true;
  }
```

src/bundler.ts
```diff
@@ -15,10 +22,30 @@ export async function isGemOutdated(): Promise<boolean> {
   return false;
 }
 
+export async function addGem(): Promise<void> {
+  await execInPath("bundle add ruby-lsp --group=development && bundle install");
```
st0012, line 26: We can use `--require=false` here so it'd be added with `:require => false`.
(I don't like the old hash syntax but it's still better than nothing imo)

src/status.ts
```diff
@@ -0,0 +1,278 @@
...
+      status.command = {
+        title: "Run bundle install",
+        command: commandId,
+      };
+      return true;
+    }
+    return false;
+  }
+
+  public async addMissingGem(): Promise<boolean> {
```
dirceu, line 117: Not a blocker as this is a matter of preference, but both `addMissingGem` and `installGems` could use an [early return](https://gomakethings.com/the-early-return-pattern-in-javascript/) to avoid having most of the function body be inside of an `if` statement.

dirceu (approved):
Thank you for your hard work on this!

When I click on `{}` -> `Stop server` (or `Start server`, if it's stopped), it opens the command palette with the following options:

Screenshot 2023-01-31 at 1 14 38 PM

Is there any way to do the thing **directly**?

If not, I suggest changing the text from "Ruby LSP: Running... <Start/stop> server" to something like "Ruby LSP: Running... Manage".

dirceu:
> Would it be a good idea to also put the option that makes the most sense at the top of the options? (e.g., if the server is running, put "Stop server" as the first option)

Great idea, yes! Then it's just click -> hit enter

adisonlampert:
@dirceu Got it working! I realized you can pass arguments into commands so I also moved every `registerCommand` call into the `registerCommands` function to clean up the code a bit and make sure `registerCommand` is only be called one time for each command 😃

PR #336, opened 2022-12-28
Add "none" as option for Ruby Version Manager

### LINKED ISSUE #335: Allow NOT to use a Ruby Version Manager
Hi,

we always develop within VS Code Remote Containers.
This means we use our production-like Dockerfile to install ruby (or specifically, we use a ruby image)

This means, we do not have the need to use a ruby version manager, as each local development environment is completely isolated.

It would be great to have an option "none" for `rubyLsp.rubyVersionManager`, which would just start whatever ruby version is installed and loaded in the `$PATH`

tmaier:
Fixes #335

Allows to use the ruby version provided within the system. This is especially useful for docker-based local dev environments.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -36,7 +36,6 @@ export class Ruby {
           await this.activate("rvm-auto-ruby");
           break;
         case "none":
-          await this.activate("ruby");
           break;
         default:
           await this.activateShadowenv();
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -35,6 +35,9 @@ export class Ruby {
         case "rvm":
           await this.activate("rvm-auto-ruby");
           break;
+        case "none":
+          await this.activate("ruby");
+          break;
         default:
           await this.activateShadowenv();
           break;
--- a/package.json
+++ b/package.json
@@ -128,7 +128,8 @@
             "chruby",
             "asdf",
             "rbenv",
-            "rvm"
+            "rvm",
+            "none"
           ],
           "default": "shadowenv"
         },


tmaier:
I just signed the CLA at https://cla.shopify.com. Please re-reun the CLA Job

src/ruby.ts
```diff
@@ -35,6 +35,9 @@ export class Ruby {
         case "rvm":
           await this.activate("rvm-auto-ruby");
           break;
+        case "none":
```
andyw8, line 38: 🤔 @vinistock I wonder if it's more coherent to make this the default, even though it means an additional step is needed for Shopify staff using `shadowenv`.

st0012: I agree that `shadowenv` is not the best default for most users in the community.

But this will likely need a coordinated change with Spin/dev to make sure we don't break Shopify devs' development experience?

paracycle: IMO the default should be "auto" where we try to discover the version manager (or the lack of it) automatically based on some search order that starts with "shadowenv" and has "none" as the ultimate fallback. If this doesn't work for specific setups, folks can still explicitly set their version manager.

src/ruby.ts
```diff
@@ -35,6 +35,9 @@ export class Ruby {
...
+          await this.activate("ruby");
```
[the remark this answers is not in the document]
vinistock: Yes, `activate` will use Ruby to extract the current environment variables and inject that into the NodeJS process where the extension runs.

But if we're not using any version manager (i.e.: `none`), then the environment variables we extract will always match the ones we already have in the NodeJS process, no?

The steps are
1. Use a version manager like chruby to change the environment variables related to Ruby (e.g.: `GEM_PATH`)
2. Print the environment as JSON
3. Grab that environment and inject it into NodeJS

When we're not using a version manager, there's no step 1, which is why I don't think there's any difference in doing 2 and 3.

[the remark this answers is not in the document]
tmaier: Done

vinistock (approved):

PR #349, opened 2023-01-11
Use interactive shell for activating Ruby

vinistock:
Instead of sourcing configuration files manually, we can use the `-i` flag to spawn an interactive shell, which does the same automatically. We only need some adjustments in the command string to make it work.

Additionally, I moved the delay to shadowenv since it's the only one that needs it and added a missing `await`.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -37,12 +37,12 @@ export class Ruby {
           break;
         default:
           await this.activateShadowenv();
+          await this.delay(500);
           break;
       }
 
-      await this.delay(500);
       await this.displayRubyVersion();
-      this.checkYjit();
+      await this.checkYjit();
     } catch (error: any) {
       vscode.window.showErrorMessage(
         `Failed to activate ${this.versionManager} environment: ${error.message}`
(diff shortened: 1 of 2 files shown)


src/ruby.ts
```diff
@@ -62,27 +62,10 @@ export class Ruby {
...
-      default:
-        shellProfilePath = `${home}/.bashrc`;
-        break;
-    }
-
     const result = await asyncExec(
-      `source ${shellProfilePath} > /dev/null 2>&1 && ${ruby} -rjson -e "puts JSON.dump(ENV.to_h)"`,
-      { shell, cwd: this.workingFolder }
+      // eslint-disable-next-line no-process-env
+      `${process.env.SHELL} -lic '${ruby} -rjson -e "puts JSON.dump(ENV.to_h)"'`,
```
andyw8, line 67: What's the `-c` flag? I don't see it listed for `zsh -h`.

vinistock: It's used to execute a single command.

andyw8: Ah, it's in the `man` page, I wonder why it's missing from the help.

```
> INVOCATION
       The following flags are interpreted by the shell when invoked to determine where the shell will read commands from:

       -c     Take the first argument as a command to execute, rather than reading commands from a script or standard input.  If any further arguments are given, the first one is assigned to $0, rather than being used as a
              positional parameter.
```

Is it safe to assume all three of these are universal to all the shells we support?

vinistock: Yes, the three flags exist in bash, zsh and fish (and they do the same thing).

PR #383, opened 2023-02-08
Add experimental features configuration to enable all features at once

vinistock:
By default, our new features should all be `enabled: false` in our configurations so that we can test them out and catch bugs before enabling them for everybody.

However, there are always developers who are willing to beta test features and report issues and we should make it easy for them to do so.

This PR adds a configuration for `enableExperimentalFeatures`, which enables all features regardless of whether they are still under development or not.


st0012 (approved):

adisonlampert (approved):
Is this something we might want to allow people to toggle on and off as a language status item? 👀

vinistock:
@adisonlampert that's a good call. I added a language status item for enabling experimental features.

I also fixed an issue with disabling and enabling YJIT. When we invoke `update`, the default is to change workspace settings, which changes the `.vscode/settings.json` file, but we actually want global user settings. The fix is just to pass the last two arguments to `update` as `true` and `true`.

PR #386, opened 2023-02-10
Delete GC tuning variables from the environment

vinistock:
Sometimes developers may use Ruby's GC tuning environment variables to calibrate their application's performance. However, this is often not adequate for the Ruby LSP and usually results in spending an insane amount of time doing garbage collection.

There might be an opportunity to tune these specifically for the Ruby LSP to output better performance, but for the time being we should at least remove those and use defaults, which greatly improves performance where GC has been tuned.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -222,6 +222,19 @@ export default class Client {
     const env = process.env;
     const useYjit = vscode.workspace.getConfiguration("rubyLsp").get("yjit");
 
+    [
+      "RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR",
+      "RUBY_GC_HEAP_FREE_SLOTS",
+      "RUBY_GC_HEAP_SLOTS_GROWTH_FACTOR",
+      "RUBY_GLOBAL_METHOD_CACHE_SIZE",
+      "RUBY_GC_MALLOC_LIMIT",
+      "RUBY_GC_HEAP_GROWTH_MAX_SLOTS",
+      "RUBY_GC_OLDMALLOC_LIMIT",
+      "RUBY_GC_HEAP_INIT_SLOTS",
+    ].forEach((key) => {
+      delete env[key];
+    });
+
     if (!this.ruby.rubyVersion) {
       return env;
     }


st0012 (approved):
I think we may want to delete `RUBYOPT` too as it doesn't make sense for the LSP server to take it anyway. 
For example, `ruby/debug` users could inject `RUBYOPT` into their shell config [following the official recommendation](https://github.com/ruby/debug/blob/95df2c5e1567c70f95ad8d134a088f5a99883141/lib/debug/client.rb#L62-L70). This means `ruby-lsp` would require that `prelude.rb` as well. Even though in this case loading that file shouldn't break `ruby-lsp`, it still highlights the risk of taking `RUBYOPT`.

vinistock:
I agree, but that might require a more complex check. If I'm not mistaken bundler itself adds some requires using `RUBYOPT` and getting rid of those could be problematic. I created an issue to investigate it separately https://github.com/Shopify/vscode-ruby-lsp/issues/387.

src/client.ts
```diff
@@ -222,6 +222,19 @@ export default class Client {
...
 
+    [
+      "RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR",
+      "RUBY_GC_HEAP_FREE_SLOTS",
+      "RUBY_GC_HEAP_SLOTS_GROWTH_FACTOR",
+      "RUBY_GLOBAL_METHOD_CACHE_SIZE",
+      "RUBY_GC_MALLOC_LIMIT",
+      "RUBY_GC_HEAP_GROWTH_MAX_SLOTS",
+      "RUBY_GC_OLDMALLOC_LIMIT",
+      "RUBY_GC_HEAP_INIT_SLOTS",
```
andyw8, line 233: Other than looking at the Ruby source, is there somewhere that documents all the available variables? Google isn't returning anything useful.

casperisfine: The most sensible thing might be to remove all the env var that start with `RUBY_GC_*`, since we introduced new ones in Ruby 3.3.0-dev already, it would be more future proof.

andyw8: Makes sense: https://github.com/Shopify/vscode-ruby-lsp/issues/388

vinistock: https://github.com/Shopify/vscode-ruby-lsp/pull/389

Issue #409, opened 2023-02-26
Option to enable logging in output


roelandmoors:
If I open the output of 'Ruby LSP' in vscode, I only see something like this:
```
Starting Ruby LSP...
Ruby LSP is ready
```
Is there a way to see more? For example what Ruby version is found or what events are happening.

The format on save is working, but I don't see linting errors, snippets are not working, ...
I'm not sure how to look for possible errors.

I have these settings:

```
  "editor.formatOnSave": true,
  "[ruby]": {
    "editor.defaultFormatter": "Shopify.ruby-lsp"
  },
  "rubyLsp.rubyVersionManager": "rbenv",
  "rubyLsp.enabledFeatures": {
    "codeActions": true,
    "diagnostics": true,
    "documentHighlights": true,
    "documentLink": true,
    "documentSymbols": true,
    "foldingRanges": true,
    "formatting": true,
    "hover": true,
    "inlayHint": true,
    "onTypeFormatting": true,
    "selectionRanges": true,
    "semanticHighlighting": true,
    "completion": true
  }
```

iTerm2 or the vscode terminal give the same results:
```
> ruby --version
ruby 3.2.1 (2023-02-08 revision 31819e82c8) [arm64-darwin22]
> env | grep SHELL
SHELL=/bin/zsh
RBENV_SHELL=zsh
```

I have this in .zshrc:
`eval "$(rbenv init - zsh)"`

vinistock:
You can enable server tracing by setting this VS Code configuration
```json
"ruby lsp.trace.server": "messages"
```

In respect to what Ruby version is being used, that information is displayed in the language status item.
1. Open any Ruby file
2. In the status bar, immediately to the left of the language mode `Ruby`, there should be a pair of braces
3. Click or hover on the braces to see information about the server, which should contain the Ruby version

Are you using RuboCop? Diagnostics are currently only supported for projects using RuboCop.

roelandmoors:
The logging works. I also see the correct Ruby version.

RuboCop is also working now, I had it installed globally, but not in the bundle.
`bundle add rubocop rubocop-rails rubocop-performance`

This feature doesn't seem to work yet:
https://shopify.github.io/ruby-lsp/RubyLsp/Requests/OnTypeFormatting.html

But the original request for a logging option is solved. Thanks!

vinistock:
About on type formatting, do you have it enabled in VS Code? Otherwise, it won't send requests to the language server.
```json
[...]

PR #411, opened 2023-02-27
Document how to enable logging

andyw8:
(in response to https://github.com/Shopify/vscode-ruby-lsp/issues/409)


vinistock (commented):
Let's include all possible values, like documented in the [ruby-lsp](https://github.com/Shopify/ruby-lsp#debugging).

andyw8:
Updated. I'll update ruby-lsp to point to here once this is merged.

vinistock (approved):

PR #413, opened 2023-02-28
Fix ZSH string interpolation issue

fnordfish:
The `#{}` string interpolation might cause the following issue.

```
Failed to activate chruby environment: Command failed: /bin/zsh -lic 'chruby-exec "ruby 3.0.2" -- ruby --disable-gems -rjson -e "puts %Q{RUBY_ENV_ACTIVATE#{JSON.dump(ENV.to_h)}RUBY_ENV_ACTIVATE}"'
(eval):1: can't change option: zle
zsh:1: no matches found: puts %Q{RUBY_ENV_ACTIVATE#{JSON.dump(ENV.to_h)}RUBY_ENV_ACTIVATE}
```

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -63,7 +63,7 @@ export class Ruby {
   private async activate(ruby: string) {
     const result = await asyncExec(
       // eslint-disable-next-line no-process-env
-      `${process.env.SHELL} -lic '${ruby} --disable-gems -rjson -e "puts %Q{RUBY_ENV_ACTIVATE#{JSON.dump(ENV.to_h)}RUBY_ENV_ACTIVATE}"'`,
+      `${process.env.SHELL} -lic '${ruby} --disable-gems -rjson -e "printf %{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump(ENV.to_h)"'`,
       { cwd: this.workingFolder }
     );
 


fnordfish:
cla signed

andyw8:
@fnordfish are you using ohmyzsh, or any other addons/extensions?

fnordfish:
Yes. I'm using oh-my-zsh.
I would really like to investigate further, but I am on a computer-less vacation :/

I'm also using iTerm2 (latest beta) 
I did try to open a blank bash (from my ZSH session) with weirdly the same effect.

aergonaut:
I also observed this issue when I tried to use the extension the first time. I tracked it down to ZSH's `EXTENDED_GLOB` option. When it is turned on, it gives special meaning to some characters, including `#`. I had `setopt EXTENDED_GLOB` in my `.zshrc`. When I removed that, the issue disappeared.

vinistock:
Thanks for the info! In that case, this seems like the right way to go to prevent it from failing in zsh. Two things
- Let's add parenthesis to the printf invocation, it's a bit hard to read already (e.g.: `printf(%{...},  JSON.dump(...))`)
- It doesn't seem like the CLA was signed. Can you double-check it?

fnordfish:
Amazing what you can do on an iPad :)

andyw8:
@vinistock should also apply this in https://github.com/Shopify/vscode-ruby-lsp/blob/main/src/ruby.ts ?

vinistock (approved):

PR #420, opened 2023-02-28
Allow restart command to fail gracefully

vinistock:
Closes Shopify/ruby-lsp#1483

If we let commands raise, VS Code won't let you use them again without reloading the window completely. That's not super helpful and we should rather catch the error and log it in the output channel.

Also, as accurately pointed out by Ufuk https://github.com/Shopify/vscode-ruby-lsp/pull/385#discussion_r1101846496, declaring a function as `async` and then only doing `await` inside of them has no difference when compared to just returning the function straight away. The behaviour will depend on whether the consumer of the function is awaiting or not (in this case VS Code). So I switched to a less verbose implementation.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -26,13 +26,13 @@ export default class Client {
   private telemetry: Telemetry;
   private ruby: Ruby;
   private statusItem: StatusItem;
+  private outputChannel = vscode.window.createOutputChannel(LSP_NAME);
 
   constructor(
     context: vscode.ExtensionContext,
     telemetry: Telemetry,
     ruby: Ruby
   ) {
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -153,21 +153,22 @@ export default class Client {
   }
 
   async restart() {
-    await this.stop();
-    await this.start();
+    try {
+      await this.stop();
+      await this.start();
+    } catch (error) {
+      this.outputChannel.appendLine(`Error restarting the server: ${error}`);
```
andyw8, line 160: Would `error.message` give more readable output?

PR #429, opened 2023-03-07
Avoid mutating the environment with RUBYOPT

vinistock:
If we simply assign `env = process.env`, then when we change keys in `env` we're mutating `process.env` as well - which leaks across different extensions.

This means that we append `RUBYOPT=--yjit` in the process environment and it ends up applying to other extensions that spawn Ruby processes, which may lead to issues.

The fix is to simply create a clone of the environment, which we can do by destructuring it inside a new object.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "ruby-lsp",
   "displayName": "Ruby LSP",
   "description": "VS Code plugin for connecting with the Ruby LSP",
-  "version": "0.1.3",
+  "version": "0.1.4",
   "publisher": "Shopify",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "ruby-lsp",
   "displayName": "Ruby LSP",
   "description": "VS Code plugin for connecting with the Ruby LSP",
-  "version": "0.1.3",
+  "version": "0.1.4",
```
KaanOzkan, line 5: I'm assuming the version bump is intentional. Maybe could be in its own commit.

st0012 (approved):

PR #437, opened 2023-03-09
Introduce `auto` option for version manager

### LINKED ISSUE #368: Introduce "auto" to `rubyVersionManager`'s options
As proposed in https://github.com/Shopify/vscode-ruby-lsp/pull/336/files#r1063406665

> IMO the default should be "auto" where we try to discover the version manager (or the lack of it) automatically based on some search order that starts with "shadowenv" and has "none" as the ultimate fallback. If this doesn't work for specific setups, folks can still explicitly set their version manager.

vinistock:
Closes #368

Introduce `auto` as the default option for version manager, which tries to guess which manager to use automatically. The logic is
- For shadowenv, check if the `.shadowenv.d` exists (otherwise the project hasn't initiated their shadowenv folder)
- For all others, verify if running `tool --version` succeeds or fails

Question: this implementation shows an error if `auto` failed to find a version manager, prompting the user to configure it. Should we set it to `none` instead?

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,16 +36,23 @@ following configuration (definition of all available values can be found in the
 
 #### Ruby version managers
 
-To boot the server properly, the Ruby version manager must be configured, which defaults to using shadowenv.
-Check the [package.json](https://github.com/Shopify/vscode-ruby-lsp/blob/main/package.json) for currently supported
-managers.
+To boot the server properly, the Ruby LSP uses a version manager to activate the right environment variables that point
+Bundler to the Ruby and gem paths. This is especially necessary when switching between projects that use different Ruby
+versions - since those paths change and need to be reactivated.
 
-The Ruby LSP uses the Ruby version manager to reactivate the required Ruby version and to select the correct gem and
-Bundler paths, which is necessary when switching between projects using different Ruby versions. Use the extension's
-`rubyLsp.rubyVersionManager` setting to control which version manager is used.
+By default, the Ruby LSP will attempt to automatically determine which version manager it should use, checking which ones are available
+(`auto` option). If that fails, then the version manager must be manually configured in VS Code's settings.
 
 ```jsonc
-"rubyLsp.rubyVersionManager": "chruby" // The handle for the version manager (e.g.: chruby, shadowenv)
+// Available options are
+// "auto" (select version manager automatically)
+// "none" (do not use a version manager)
+// "asdf"
+// "chruby"
+// "rbenv"
+// "rvm"
+// "shadowenv"
+"rubyLsp.rubyVersionManager": "chruby"
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -36,16 +36,23 @@ following configuration (definition of all available values can be found in the
 
 #### Ruby version managers
 
-To boot the server properly, the Ruby version manager must be configured, which defaults to using shadowenv.
-Check the [package.json](https://github.com/Shopify/vscode-ruby-lsp/blob/main/package.json) for currently supported
-managers.
+To boot the server properly, the Ruby LSP uses version managers to activate the right environment variables that point
+bundler to the Ruby and gem paths. This is especially necessary when switching between projects that use different Ruby
```
andyw8, line 40:
```suggestion
Bundler to the Ruby and gem paths. This is especially necessary when switching between projects that use different Ruby
```

README.md
```diff
@@ -36,16 +36,23 @@ following configuration (definition of all available values can be found in the
...
+versions - since those paths change and need to be reactivated.
 
-The Ruby LSP uses the Ruby version manager to reactivate the required Ruby version and to select the correct gem and
-Bundler paths, which is necessary when switching between projects using different Ruby versions. Use the extension's
-`rubyLsp.rubyVersionManager` setting to control which version manager is used.
+By default, the Ruby LSP will attempt to guess which version manager it should use, checking which ones are available
```
andyw8, line 43:
```suggestion
By default, the Ruby LSP will attempt to automatically determine which version manager it should use, checking which ones are available
```

src/ruby.ts
```diff
@@ -117,6 +130,46 @@ export class Ruby {
...
+      const exists = await this.toolExists(tool);
+
+      if (exists) {
+        this.versionManager = tool;
+        return;
+      }
+    }
+
+    await vscode.window.showErrorMessage(
+      "Failed to discover Ruby version manager automatically. \
```
andyw8, line 158:
```suggestion
      "Failed to discover a Ruby version manager automatically. \
```

andyw8:
> Question: this implementation shows an error if auto failed to find a version manager, prompting the user to configure it. Should we set it to none instead?

I think it's a rare that `none` would the desired setting, and it's better to inform the user.

paracycle (commented):
Thank you, this looks great!

My recommendation on the original issue was to fallback to `none` if we can't find any version managers. There is hardly anything the user can do if we show an error, and if the selection of `none` was wrong, they can always pick a better one from the UI.
> ... "none" as the ultimate fallback

vinistock:
Reflecting more about the none vs error thing, I agree with using none. We have received feedback in the past about over notifying users, so we should resist the urge to reach for that.

PR #438, opened 2023-03-09
Refactor status items into separate classes

vinistock:
This is a suggestion refactor on our status items. While working on Shopify/ruby-lsp#1542, I noticed the `Status` class is getting more and more complex and it's difficult to see what's related to a single item.

This PR
- Refactors all status items as separate classes that all inherit from `Status`. They are all initialized through their constructors, implement a `registerCommand` function in case they need to register and implement a `refresh` method in case their information changes on a restart
- Adds tests for all different status item classes

### Drawbacks

The weird thing about this approach is that `refresh` will always need to accept objects required to update the items - even if some of them don't require the same objects. For example, `status` is only pertinent to `ServerStatus`, but all of them receive it as an argument.

Still I think it's significantly easier to go over which statuses exist.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -13,7 +13,7 @@ import {
 
 import { Telemetry } from "./telemetry";
 import { Ruby } from "./ruby";
-import { StatusItem, Command } from "./status";
+import { StatusItems, Command, ServerState, ClientInterface } from "./status";
 
 const LSP_NAME = "Ruby LSP";
 const asyncExec = promisify(exec);
@@ -22,15 +22,16 @@ interface EnabledFeatures {
   [key: string]: boolean;
 }
 
-export default class Client {
+export default class Client implements ClientInterface {
   private client: LanguageClient | undefined;
-  private context: vscode.ExtensionContext;
+  private _context: vscode.ExtensionContext;
   private workingFolder: string;
   private clientOptions: LanguageClientOptions;
   private telemetry: Telemetry;
-  private ruby: Ruby;
-  private statusItem: StatusItem;
+  private _ruby: Ruby;
+  private statusItems: StatusItems;
   private outputChannel = vscode.window.createOutputChannel(LSP_NAME);
+  private _state: ServerState = ServerState.Starting;
 
   constructor(
     context: vscode.ExtensionContext,
@@ -39,7 +40,7 @@ export default class Client {
   ) {
     this.workingFolder = vscode.workspace.workspaceFolders![0].uri.fsPath;
     this.telemetry = telemetry;
-    this.ruby = ruby;
+    this._ruby = ruby;
 
     this.clientOptions = {
       documentSelector: [{ scheme: "file", language: "ruby" }],
@@ -107,13 +108,16 @@ export default class Client {
       },
     };
 
-    this.context = context;
-    this.statusItem = new StatusItem(this.context, this.ruby);
+    this._context = context;
+    this.statusItems = new StatusItems(this);
     this.registerCommands();
     this.registerAutoRestarts();
   }
 
   async start() {
+    this._state = ServerState.Starting;
+    this.statusItems.refresh();
+
     await vscode.window.withProgress(
       {
         location: vscode.ProgressLocation.Notification,
@@ -150,21 +154,25 @@ export default class Client {
       this.clientOptions
     );
 
-    await this.statusItem.refresh(Command.Start, this.ruby);
-
     this.client.onTelemetry((event) =>
       this.telemetry.sendEvent({
         ...event,
-        rubyVersion: this.ruby.rubyVersion,
-        yjitEnabled: this.ruby.yjitEnabled,
+        rubyVersion: this._ruby.rubyVersion,
+        yjitEnabled: this._ruby.yjitEnabled,
       })
     );
+
     await this.client.start();
+
+    this._state = ServerState.Running;
+    this.statusItems.refresh();
   }
 
   async stop(): Promise<void> {
     if (this.client) {
-      await this.statusItem.refresh(Command.Stop, this.ruby);
+      this._state = ServerState.Stopped;
+      this.statusItems.refresh();
+
       return this.client.stop();
     }
   }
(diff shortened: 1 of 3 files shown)


src/status.ts
```diff
@@ -23,123 +23,113 @@ const STARTED_SERVER_OPTIONS = [
...
-  private experimentalFeaturesStatus: vscode.LanguageStatusItem;
+abstract class Status {
+  public item: vscode.LanguageStatusItem;
+  protected context: vscode.ExtensionContext;
 
-  constructor(context: vscode.ExtensionContext, ruby: Ruby) {
-    this.context = context;
-    this.ruby = ruby;
-    this.selector = {
+  constructor(id: string, context: vscode.ExtensionContext, _ruby: Ruby) {
```
paracycle, line 30: It feels like a better API might be:
```suggestion
  constructor(id: string, client: { context: vscode.ExtensionContext, ruby: Ruby }) {
```
so that status items store references to the `client` and when `refresh` is called, no arguments need to be passed, they can each access the relevant value from the `client`.

Since the typing for `client` is structural, you can easily replace that in tests with a fake client instance that has the same two properties.

src/status.ts
```diff
@@ -23,123 +23,113 @@ const STARTED_SERVER_OPTIONS = [
...
 
-export class StatusItem {
-  private context: vscode.ExtensionContext;
-  private ruby: Ruby;
-  private selector: vscode.DocumentSelector;
-  private serverStatus: vscode.LanguageStatusItem;
-  private yjitStatus: vscode.LanguageStatusItem;
-  private rubyVersionStatus: vscode.LanguageStatusItem;
-  private experimentalFeaturesStatus: vscode.LanguageStatusItem;
+abstract class Status {
```
paracycle, line 26: I think I like the `StatusItem` name better for this, tbh. Especially since the collection name is `StatusItems`.

paracycle (commented):
I like this refactor a lot.

src/status.ts
vinistock: I like this! My original idea was to pass the actual client object to have easy access to the state. But unfortunately, that introduces a circular dependency and TypeScript won't have it.

src/test/suite/status.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      assert.strictEqual(status.item.command?.title, "Disable");
+      assert.strictEqual(status.item.command?.command, Command.ToggleYjit);
+      assert.strictEqual(context.subscriptions.length, 1);
+    });
+
+    test("Refresh updates whether it's disabled or enabled", async () => {
+      assert.strictEqual(status.item.text, "YJIT enabled");
+
+      status.refresh(Command.ToggleYjit, { supportsYjit: false } as Ruby);
+      assert.strictEqual(status.item.text, "YJIT disabled");
```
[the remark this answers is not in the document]
vinistock: Added a separate example for trying to enable through configuration when it's not supported 👍

src/status.ts
```diff
@@ -23,123 +23,120 @@ const STARTED_SERVER_OPTIONS = [
...
+  ) {
+    this.item = vscode.languages.createLanguageStatusItem(id, {
       scheme: "file",
       language: "ruby",
+    });
+    this.context = client.context;
+    this.registerCommand();
+  }
+
+  abstract refresh(status: Command, ruby: Ruby): void;
```
paracycle, line 42: Do we really need the `ruby` parameter anymore, seen as we already have that being passed to the constructor? I think just a `status` should be enough:
```suggestion
  abstract refresh(status: Command): void;
```

src/client.ts
```diff
@@ -108,7 +108,10 @@ export default class Client {
     };
 
     this.context = context;
-    this.statusItem = new StatusItem(this.context, this.ruby);
+    this.statusItems = new StatusItems({
+      context: this.context,
+      ruby: this.ruby,
+    });
```
paracycle, line 114: I was thinking that this could be more flexible if we were passing the `client` itself, so that the status item could refer to the current value of `context` and `ruby` at any given point. I guess that means we might need to make (at least the getters of) `ruby` and `context` public.

So, we can do something like [this](https://www.typescriptlang.org/play?#code/MYGwhgzhAEDC0G8CwAoa0AOAnAlgNzABcBTaAfRGIDsBzQgC2gF5oAGAblVXQwFcAjEDmDQaxQtEq0GACgCUibumhZxvLFWgMcEAHQVqdepzTQAvkuz4ipCOMmHZBEL2ILkp9Nr0HpjFs6uJugWKKGolBIiLFTEAO5w8ibAulJGqClpDMzQACwATOxAA):
```ts
class Client {
  private _ruby: Ruby

  public get ruby() {
    return this._ruby;
  }

  private set ruby(ruby: Ruby) {
    this._ruby = ruby;
  }
}

client = new Client();

client.ruby # should work
client.ruby = new Ruby() # should fail
```

vinistock: We can't pass the client itself. TypeScript won't accept a `Client` for `{ context: ..., ruby: ... }`. And we can't import `Client` in `StatusItem` because it causes a circular dependency.

paracycle: It only rejects passing a `client` as `{ context: ..., ruby: ... }` since those getters are currently private on `Client`. If we fix that, it should work.

vinistock: Perfect! I used that and changed refresh to accept the Client interface as well. Much better.

src/status.ts
```diff
@@ -23,123 +30,124 @@ const STARTED_SERVER_OPTIONS = [
...
-  private serverStatus: vscode.LanguageStatusItem;
-  private yjitStatus: vscode.LanguageStatusItem;
-  private rubyVersionStatus: vscode.LanguageStatusItem;
-  private experimentalFeaturesStatus: vscode.LanguageStatusItem;
-
-  constructor(context: vscode.ExtensionContext, ruby: Ruby) {
-    this.context = context;
-    this.ruby = ruby;
-    this.selector = {
+export interface Client {
```
paracycle, line 33: It would be great if we could rename this to `ClientInterface` (or similar) and then used it in `client.ts` as `class Client implements ClientInterface`. That would ensure that the interface cannot go out of sync and makes the contract more explicit.

src/status.ts
```diff
@@ -23,123 +30,124 @@ const STARTED_SERVER_OPTIONS = [
...
+  constructor(id: string, client: Client) {
+    this.item = vscode.languages.createLanguageStatusItem(id, {
       scheme: "file",
       language: "ruby",
+    });
+    this.context = client.context;
+    this.registerCommand();
+  }
+
+  abstract refresh(client: Client): void;
```
paracycle, line 52: Do we still need the client here, if we were to store the `client` inside `constructor` and used `this.client` inside the `refresh` method to access properties?

src/status.ts
vinistock: Great idea.

src/status.ts
vinistock: We don't indeed. Refresh takes no arguments now

paracycle (approved):
This looks amazing 🤩

PR #440, opened 2023-03-10
Move environment manipulation completely to the Ruby object

vinistock:
The issue in #429 got me thinking about how we also mutate the environment during activation and if we manipulating it in the client made sense.

I think all environment manipulation should be contained in the `Ruby` class - since we're interested in the Ruby environment after all. This PR
- moves all env related behaviour inside `Ruby`
- stops mutating `process.env` completely
- adds a few tests, limited to what we can verify without a version manager installed
- adds setup-ruby to CI, since we need some Ruby to be installed in order to check for the information

### DIFF
--- a/src/test/suite/ruby.test.ts
+++ b/src/test/suite/ruby.test.ts
@@ -1,0 +1,59 @@
+import * as assert from "assert";
+
+import { before, after } from "mocha";
+import * as vscode from "vscode";
+
+import { Ruby } from "../../ruby";
+
+suite("Ruby environment activation", () => {
+  let ruby: Ruby;
+  const configuration = vscode.workspace.getConfiguration("rubyLsp");
+  const currentManager = configuration.get("rubyVersionManager")!;
+
+  before(async () => {
+    await configuration.update("rubyVersionManager", "none", true, true);
+  });
+
+  after(async () => {
+    await configuration.update(
+      "rubyVersionManager",
+      currentManager,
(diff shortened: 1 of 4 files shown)


vinistock:
I think it's because I used `afterEach` instead of `after` for restoring your configured version manager. Can you give it another try?

src/test/suite/ruby.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    ruby = new Ruby("fake/some/project");
+    await ruby.activateRuby();
+
+    const [major, minor, _patch] = ruby.rubyVersion!.split(".").map(Number);
+    const supportsYjit = ruby.yjitEnabled && [major, minor] >= [3, 2];
+
+    assert.ok(ruby.rubyVersion, "Expected Ruby version to be set");
+    assert.strictEqual(
+      ruby.supportsYjit,
+      supportsYjit,
```
st0012, line 36: If we're only testing against Ruby 3.2, isn't this always `true`?
If we want to cover other scenarios, perhaps we should have a Ruby version matrix?

st0012:
@vinistock Tests now all pass locally. Thanks 👍

st0012 (approved):

src/test/suite/ruby.test.ts
st0012: Thanks for simplifying it 👍

PR #443, opened 2023-03-13
Only bundle install when necessary

vinistock:
Closes #439

Currently, we copy the top level `Gemfile.lock` on every activation, but that's wasteful and runs bundle install way more than necessary. This PR makes it so that we only copy the `Gemfile.lock` and run bundle install if
- There has been more than a we haven't done so. This way we update the server gem automatically daily
- The top level `Gemfile.lock` was modified more recently than the custom `Gemfile.lock`
- Bundle check fails. This could happen if people uninstall gems manually

I also added a command to force update the server, which could be useful for getting bug fixes out quicker.


src/client.ts
```diff
@@ -17,6 +17,7 @@ import { StatusItems, Command, ServerState, ClientInterface } from "./status";
 
 const LSP_NAME = "Ruby LSP";
 const asyncExec = promisify(exec);
+const ONE_DAY = 24 * 60 * 60 * 1000;
```
andyw8, line 20:
```suggestion
const ONE_DAY_IN_MS = 24 * 60 * 60 * 1000;
```

src/client.ts
```diff
@@ -403,4 +399,87 @@ export default class Client implements ClientInterface {
...
     }
   }
+
+  private async gemsAreInstalled(customGemfilePath: string): Promise<boolean> {
+    try {
+      await asyncExec(`BUNDLE_GEMFILE=${customGemfilePath} bundle check`, {
+        cwd: this.workingFolder,
+      });
+      return true;
+    } catch {
```
andyw8, line 409: Worth logging the exception here?

src/client.ts
```diff
@@ -403,4 +399,87 @@ export default class Client implements ClientInterface {
...
+    if (!fs.existsSync(gemfileLockPath)) {
+      return false;
+    }
+
+    // If there's no custom Gemfile.lock, then it hasn't been created yet and we must sync
+    if (!fs.existsSync(customGemfileLockPath)) {
+      return true;
+    }
+
+    // If the last motified time of the top level Gemfile.lock is greater than the custom Gemfile.lock, then changes
```
andyw8, line 428:
```suggestion
    // If the last modified time of the top level Gemfile.lock is greater than the custom Gemfile.lock, then changes
```

package.json
```diff
@@ -35,6 +35,10 @@
       {
         "command": "rubyLsp.stop",
         "title": "Ruby LSP: Stop"
+      },
+      {
+        "command": "rubyLsp.update",
+        "title": "Ruby LSP: Update server gem"
```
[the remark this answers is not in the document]
andyw8: Or even just 'update'? Since the `gem` aspect is kinda hidden from the user now.

andyw8 (approved):

src/client.ts
```diff
@@ -271,19 +260,26 @@ export default class Client implements ClientInterface {
...
-    // RuboCop and related extensions. Because we do this in every initialization, we always use the latest version of
-    // the Ruby LSP
-    if (fs.existsSync(path.join(this.workingFolder, "Gemfile.lock"))) {
-      fs.cpSync(
-        path.join(this.workingFolder, "Gemfile.lock"),
-        path.join(this.workingFolder, ".ruby-lsp", "Gemfile.lock")
-      );
-    }
+    const lastUpdatedAt: number =
+      this._context.workspaceState.get("rubyLsp.lastBundleInstall") ?? 0;
```
adisonlampert, line 264: What is the reasoning behind setting `lastUpdateAt` to `0` if there is nothing set at `rubyLsp.lastBundleInstall` in the workspace state? We are using this variable to see if the last update check was more than a day ago: `Date.now() - lastUpdatedAt > ONE_DAY`. When there is nothing at workspace state, `lastUpdateAt = 0`, this logic would be false and we wouldn't check for an update. Is there ever a scenario where `rubyLsp.lastBundleInstall` is just never set so we never check for an update?

src/client.ts
```diff
@@ -403,4 +399,87 @@ export default class Client implements ClientInterface {
...
+        try {
+          await asyncExec(
+            `BUNDLE_GEMFILE=${customGemfilePath} bundle install`,
+            {
+              cwd: this.workingFolder,
+            }
+          );
+        } catch (error: any) {
+          // The progress dialog can't be closed by the user, so we have to guarantee that we catch errors
+          vscode.window.showErrorMessage(
```
adisonlampert, line 471: In this case, it would be a good idea to have the status item for server status be set to `Error`. I'm not sure how to trigger this error to check so is that something that's already happening or does the server status need to be updated manually?

adisonlampert (approved):

src/client.ts
```diff
@@ -271,19 +260,26 @@ export default class Client implements ClientInterface {
...
+      this.workingFolder,
+      ".ruby-lsp",
+      "Gemfile.lock"
+    );
 
-    await asyncExec(`BUNDLE_GEMFILE=${customGemfilePath} bundle install`, {
-      cwd: this.workingFolder,
-    });
+    // Copy the Gemfile.lock and install gems to get `ruby-lsp` updates if
+    // - it's been more than a day since the last time we checked for updates
```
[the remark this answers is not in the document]
vinistock: Even if we did, I'd still prefer to keep the comment to be honest.

src/client.ts
vinistock: When gems are not installed, running bundle check will return status 1 and fail, which leads us to this branch. There's not much to learn from logging here, this just indicates gems aren't installed.

src/client.ts
vinistock: Fair point! I stopped defaulting it to zero and starting checking for `undefined` instead. If it's `undefined` (we never bundle installed), then it will do it and save something in the workspace state.

src/client.ts
vinistock: It would indeed! I added a line to update the state and refresh the status items. It needs to be manually updated.

PR #445, opened 2023-03-14
Allow marking features as experimental

### LINKED ISSUE #399: Disabling diagnostics is ignored when enableExperimentalFeatures is enabled
I use a VSCode profile with the following settings: 

```
{
    "rubyLsp.rubyVersionManager": "rbenv",
    "rubyLsp.enableExperimentalFeatures": true,
    "rubyLsp.enabledFeatures": {
    
        "codeActions": true,
        "diagnostics": false, // disabled
        "documentHighlights": true,
        "documentLink": true,
        "documentSymbols": true,
        "foldingRanges": true,
        "formatting": true,
        "hover": true,
        "inlayHint": true,
        "onTypeFormatting": true,
        "selectionRanges": true,
        "semanticHighlighting": true,
        "completion": true
    },
    "git.confirmSync": false,
    "window.zoomLevel": 1
}
```

My Gemfile gets highlighted for having double-quoted strings: 
image

As soon as I set `"rubyLsp.enableExperimentalFeatures": false,` the issue disappears. 

Given this code it's actually not surprising, it's a bit odd that enabling experimental features would override configurations made afterwards: https://github.com/olivierlacan/vscode-ruby-lsp/blob/e233fb2da0c8584ce187c243a4de44f4b646ce7f/src/client.ts#L217-L220

vinistock:
Closes #399

Currently, if experimental features are enabled, we're just enabling everything. That behaviour is not very accurate and probably surprising. We need a way of marking features as experimental so that only those are enabled by the setting.

Additionally, it's the server that should decide whether a feature is experimental or not - otherwise they could be accidentally enabled by other editors.

This PR switches our mechanism to just send whether experimental features are enabled or not to the server. When returning capabilities, the server can then read this and determine if it wants to broadcast experimental feature capabilities or not.

Note: we don't have any experimental features at this time, so there's no PR in the server to accompany this one.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -49,6 +49,9 @@ export default class Client implements ClientInterface {
       revealOutputChannelOn: RevealOutputChannelOn.Never,
       initializationOptions: {
         enabledFeatures: this.listOfEnabledFeatures(),
+        experimentalFeaturesEnabled: vscode.workspace
+          .getConfiguration("rubyLsp")
+          .get("enableExperimentalFeatures"),
       },
       middleware: {
         provideOnTypeFormattingEdits: async (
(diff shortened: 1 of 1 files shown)


paracycle (approved):
Makes sense

PR #451, opened 2023-03-14
Allow users to choose formatter

vinistock:
Closes Shopify/ruby-lsp#1541

The server side is implemented in https://github.com/Shopify/ruby-lsp/pull/556.

Add configuration to allow users to configure which formatter to use. I suggest we allow the following options by default
- `auto`: the server tries to decide between RuboCop or Syntax Tree automatically (the default)
- `none`: do not use a formatter
- For all others, just the name of the formatter (e.g.: rubocop, syntax_tree...)

The option is passed as an initialization option to the server, which then decides how to format files using the given tool.

Note: in order for this to work properly, I have moved creating our `clientOptions` inside `start`. This is necessary because we need to read configurations again every time restarting the server, or else changing the formatter would never be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The tool to be used for formatting files can be configured with the following se
 
 ```jsonc
 // Available options
-//    auto: automatically detect the formatter based on the app's bundle
+//    auto: automatically detect the formatter based on the app's bundle (default)
 //    none: do not use a formatter (disables format on save and related diagnostics)
 //    all other options are the name of the formatter (e.g.: rubocop or syntax_tree)
 "rubyLsp.formatter": "auto"


README.md
```diff
@@ -59,6 +59,18 @@ To make sure that the Ruby LSP can find the version manager scripts, make sure t
 configuration script (e.g.: ~/.bashrc, ~/.zshrc) and that the SHELL environment variable is set and pointing to the
 default shell.
 
+#### Configuring a formatter
+
+The tool to be used for formatting files can be configured with the following setting.
+
+```jsonc
+// Available options
+//    auto: automatically detect the formatter based on the app's bundle
```
st0012, line 68: Do we want to mention `auto` is the default like we did in the `Ruby version managers` section?

```suggestion
//    auto: automatically detect the formatter based on the app's bundle (default)
```

st0012 (approved):
👍 Just a small suggestion

gurgeous:
I wonder if there is a bug here. My project uses standardrb (a popular wrapper around rubocop) and I think the `auto` setting wasn't detecting it. When I switched the setting to `rubocop` formatting worked as expected.

PR #454, opened 2023-03-15
Match `.ruby-version` behaviour of `chruby-auto`

paracycle:
The [`chruby-auto` script][0] in chruby will look for a `.ruby-version` file by walking up the current directory tree. This is useful when a project does not include a `.ruby-version` file, but the user has configured a default version in their home directory, or in a parent directory where they checkout all of their projects.

By default `chruby-auto` will walk up the directory tree [until it hits a directory that doesn't exist][1]. When it reads a `.ruby-version` in a directory, it only considers the version read from that file [if it is not an empty string][2]. Otherwise, it continues walking up the tree.

This commit matches that behaviour in our extension as well.

[0]: https://github.com/postmodern/chruby/blob/master/share/chruby/auto.sh
[1]: https://github.com/postmodern/chruby/blob/master/share/chruby/auto.sh#L6-L7
[2]: https://github.com/postmodern/chruby/blob/master/share/chruby/auto.sh#L9

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -1,6 +1,5 @@
 import { exec } from "child_process";
 import { promisify } from "util";
-import { readFile } from "fs/promises";
 import path from "path";
 import fs from "fs";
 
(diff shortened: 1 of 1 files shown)


vinistock (approved):

andyw8:
I haven't checked, I just thought it may be similar.

paracycle:
Isn't the linked section saying exactly that `asdf` does this automatically?
> `asdf` performs a version lookup of a tool in all `.tool-versions` files from the current working directory up to the `$HOME` directory. The lookup occurs just-in-time when you execute a tool that `asdf` manages.

PR #457, opened 2023-03-16
Update to Node 18.15.0

vinistock:
We can't get CI to pass on #456 because of Node 17. This PR just upgrades to Node 18.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
-          node-version: "17"
+          node-version: "18"
           cache: "yarn"
 
       # We need some Ruby installed for the environment activation tests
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
-          node-version: "17"
+          node-version: "18"
```
andyw8, line 15: Would it be viable to leave this out so it always uses the latest major Node release?

andyw8: (but I suppose we'd want it to match `dev.yml`).

andyw8 (approved):

.github/workflows/ci.yml
vinistock: I think we're better off using stable releases, which are the even numbers. For example, version 19 is already out, but it's an unstable release.

PR #462, opened 2023-03-17
Improve `bundle install` experience

paracycle:
The current way that we show a progress notification during `bundle install` does not give any information about what's happening to the user and the whole process is not cancellable (in case it takes too long, or something goes wrong).

By using `spawn` instead of `exec`, we can actually stream `stdout` and `stderr` to any destination we want. By coupling that with `readline` interfaces that read from `stdout` and `stderr`, we can also read that data line by line.

This allows us to:
1. Stream the stdout and stderr of the `bundle install` process to the output tab as prefixed strings.
2. Stream each line of the stdout to the progress notification window, so that it always shows the latest line of the output.
3. Allow the user to cancel the operation, which we use to cancel the `spawn` operation (via an abort signal).

As a follow up, we can also improve the error message we show after a failed `bundle install` and include a `Show Output` button, which users can click to see the `Output` panel and the Bundler error that would be in there.


vinistock (approved):
Very nice!

vinistock (approved):

PR #465, opened 2023-03-21
Show current manager when configuring

vinistock:
When selecting the version manager, we didn't inform the user in any way about which one is currently selected. This PR just adds that information as the placeholder of the quick pick.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -22,7 +22,7 @@ export class Ruby {
   public yjitEnabled?: boolean;
   public supportsYjit?: boolean;
   private workingFolder: string;
-  private versionManager?: VersionManager;
+  #versionManager?: VersionManager;
   // eslint-disable-next-line no-process-env
   private shell = process.env.SHELL;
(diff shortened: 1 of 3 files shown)


paracycle (commented):
Nice, I felt like this was missing. One question though: Does the placeholder mean that the existing manager is show as being preselected? If so, that would be nice.

Can we also add the same information to the selected Ruby status item? I think it would help to see which version manager was used to select the current Ruby.

paracycle:
Ah I guess if we used a `QuickPickItem` for each entry, we could mark the existing one as `picked`: https://code.visualstudio.com/api/references/vscode-api#QuickPickItem

vinistock:
I started showing the resolved version manager next to the Ruby version (that means if `auto` is being used, it will show which manager we discovered).

Unfortunately, the `picked` attribute only works for multi-select. It doesn't pre-select the option if we set it in a regular quick pick.

paracycle (approved):

PR #476, opened 2023-03-21
Decouple code units

paracycle:
Currently most of our extension logic revolves around the `Client` class, and it ends up orchestrating all the other functionality in the extension. However, that creates unnecessary dependencies between the `Client` unit and other units.

Instead, it should be the top level extension that manages the separate units and any interaction between said units should happen via events, as appropriate.

For that reason, this PR refactors the extension heavily to introduce:

1. A top level `Extension` class that encapsulates the creation/activation/deactivation logic for all related units.
2. A separate unit to hold all the `enum` definitions.
3. An `onStateChanged` event on the `Client` class that can be used to watch for and respond to state changes. The `StatusItems` class uses that to call the `refresh` method on individual status items as appropriate.
4. As a corollary to the previous step, individual status items don't need to hold a reference to any kind of client or context object and can simply use the object passed as event argument.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -34,6 +34,7 @@ export default class Client implements ClientInterface {
   #context: vscode.ExtensionContext;
   #ruby: Ruby;
   #state: ServerState = ServerState.Starting;
+  #onStateChangeEmitter = new vscode.EventEmitter<ClientInterface>();
 
   constructor(
     context: vscode.ExtensionContext,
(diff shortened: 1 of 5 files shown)


vinistock (commented):
I love the event emitter, the separate enums file and the new way to register commands.

However, I'm not super I'm onboard with
- The new Extension class. I'm not seeing the value and I would vote to maintain our simpler approach of just implementing stuff inside `activate` and `deactivate`
- The fact that we now need to pass `client` to `refresh` in statuses. Why is that needed again? Can't we access `this.client` for the updated information?

PR #506, opened 2023-03-29
Run entire test class in terminal

adisonlampert:
### Description

The first step of #498 (related to https://github.com/Shopify/ruby-lsp/pull/598)

Creates the `rubyLsp.runTest` VSCode command that is used in the `CodeLens` request [here](https://github.com/Shopify/ruby-lsp/pull/598/files#diff-9032a8992a4018ba0a0d29b93314fa22ef0ded7889412ccf05fd6125a36f728aR55). This command runs the entire test class in a terminal when `Run` is clicked.

code_lens

### Manual Testing

1. Debug this branch on the [`al/add-code-lens`](https://github.com/Shopify/ruby-lsp/tree/al/add-code-lens) branch of the Ruby LSP.
2. Open up a test file (ex. `test/document_test.rb`)
3. Observe there is an option to "Run" above the class declaration
4. Click "Run"
5. Observe the terminal is opened and the tests are run
6. Navigate to a non-test Ruby file. Observe the option to "Run" does not appear.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,11 @@
               "description": "Enable completion, which provides suggestions for code completion",
               "type": "boolean",
               "default": true
+            },
+            "codeLens": {
+              "description": "Enable code lens, which generates clickable text above tests to run them in the terminal",
+              "type": "boolean",
+              "default": false
             }
           },
           "default": {
@@ -132,7 +137,8 @@
             "onTypeFormatting": true,
             "selectionRanges": true,
             "semanticHighlighting": true,
-            "completion": true
+            "completion": true,
+            "codeLens": false
           }
         },
         "rubyLsp.rubyVersionManager": {
--- a/src/client.ts
+++ b/src/client.ts
@@ -34,6 +34,7 @@ export default class Client implements ClientInterface {
     | vscode.CancellationTokenSource
     | undefined;
 
+  private terminal: vscode.Terminal | undefined;
   #context: vscode.ExtensionContext;
   #ruby: Ruby;
   #state: ServerState = ServerState.Starting;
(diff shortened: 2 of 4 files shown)


vinistock:
These changes look good to me. Can we also add the code lens configuration already? But let's leave it as `false` for now. We need to add [here](https://github.com/Shopify/vscode-ruby-lsp/blob/fe9ea9dba8607f6c7d36fd9acb07c12dfc87c2e1/package.json#L116) and [here](https://github.com/Shopify/vscode-ruby-lsp/blob/fe9ea9dba8607f6c7d36fd9acb07c12dfc87c2e1/package.json#L135).

src/client.ts
```diff
@@ -235,10 +239,20 @@ export default class Client implements ClientInterface {
...
         this.updateServer.bind(this)
-      )
+      ),
+      vscode.commands.registerCommand(Command.RunTest, this.runTest.bind(this))
     );
   }
 
+  private runTest(_path: string, _name: string, command: string) {
+    if (this.terminal === undefined) {
+      this.terminal = vscode.window.createTerminal({ name: "Run test" });
```
andyw8, line 249: Is the name mandatory? I'm wondering it makes sense to call it that, since someone could choose to keep the terminal open afterwards for general usage.

package.json
```diff
@@ -117,6 +117,11 @@
               "description": "Enable completion, which provides suggestions for code completion",
               "type": "boolean",
               "default": true
+            },
+            "codeLens": {
+              "description": "Enable code lens, which generates clickable text above tests to run them in the terminal",
+              "type": "boolean",
+              "default": true
```
vinistock, line 124: Here let's use `false` as the default for now.

vinistock (commented):
You'll need to fix some tests related to the number of enabled features.

vinistock (approved):

andyw8 (approved):

PR #507, opened 2023-03-30
Also "activate" ruby if `VersionManager.none` was specified

kemchenj:
Currently if `VersionManager.none` was specified, ruby will load from system default path.

Which won't load `~/.bashrc` or `~/.zshrc` like other `VersionManager` option. For macOS, it will always use the system integrated ruby, which is not a good news for homebrew installed ruby.

So I think it might be better to also "activate" ruby when using `VersionManager.none`.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -77,6 +77,7 @@ export class Ruby {
           await this.activate("rvm-auto-ruby");
           break;
         case VersionManager.None:
+          await this.activate("ruby");
           break;
         default:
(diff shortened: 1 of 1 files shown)


vinistock:
Thank you for the detailed explanation. I believe this is indeed the right solution, but we'll need to improve these flaky tests otherwise they'll never pass. I'll see what I can do.

kemchenj:
Sure, I rebased the main branch and pushed, please have a look again.

PR #513, opened 2023-03-30
Detect `ruby-lsp` being in the `Gemfile` through `Bundler`

### LINKED ISSUE #500: Detect `ruby-lsp` being in the `Gemfile` through `Bundler`
We are currently trying to detect if the project depends on `ruby-lsp` by doing a `bundle show ruby-lsp` but that has 2 problems:

1. It would give us a false positive if there ever was a gem that had a dependency on `ruby-lsp` that was listed in the `Gemfile`, but not `ruby-lsp` itself.
2. It would fail if the user had not run `bundle install` yet.

The better way to check if the project has a dependency on `ruby-lsp` is to look at the lockfile to see if `ruby-lsp` is listed as a dependency (i.e. a direct dependency) or not.

We could do that by either running a `^  ruby-lsp \([^)]+\)$` regexp on the `Gemfile.lock` file, or by running `bundle exec ruby -e "exit 1 if Bundler.locked_gems.dependencies.key?('ruby-lsp')"` in a Ruby shell

andyw8:
Closes #500

## Tophatting

* Open a project which doesn't contain `ruby-lsp` in the Gemfile, e.g. `code-db`.
* Ensure that `.ruby-lsp/Gemfile` exists with an entry for `ruby-lsp`
* Add `ruby-lsp` to the Gemfile
* Reload the window
* Ensure that `.ruby-lsp/Gemfile` exists with a **commented out** entry for `ruby-lsp`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -352,13 +352,9 @@ export default class Client implements ClientInterface {
   }
 
   private async rubyLspIsInBundle(): Promise<boolean> {
-    // When working on the Ruby LSP itself, it's always included in the bundle
-    if (path.basename(this.workingFolder) === "ruby-lsp") {
-      return true;
-    }
-
     try {
-      // detect if ruby-lsp is a direct dependency, but ignore if it's transitive
+      // exit with an error if ruby-lsp not a dependency or is a transitive dependency.
+      // exit with success if ruby-lsp is a direct dependency.
       await asyncExec(
         `BUNDLE_GEMFILE=${path.join(
           this.workingFolder,
--- a/src/client.ts
+++ b/src/client.ts
@@ -358,12 +358,12 @@ export default class Client implements ClientInterface {
     }
 
     try {
-      // If bundle show succeeds, it means the ruby-lsp gem is a part of the bundle
+      // detect if ruby-lsp is a direct dependency, but ignore if it's transitive
       await asyncExec(
         `BUNDLE_GEMFILE=${path.join(
           this.workingFolder,
           "Gemfile"
-        )} bundle show ruby-lsp`,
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -358,20 +358,20 @@ export default class Client implements ClientInterface {
...
         {
           cwd: this.workingFolder,
           env: this.ruby.env,
         }
       );
-      return true;
-    } catch (error) {
       return false;
+    } catch (error) {
+      return true;
```
andyw8, line 374: Something is confusing here, I'm seeing a breakpoint hit for `return true`, but then it hits the `else` clause in `if await this.rubyLspIsInBundle()`, am I missing some subtlety of the async behaviour @vinistock ?

vinistock: I'm not 100% sure. But my first suggestion would be to invert the logic to make the success case `true`. That is
```
exit 1 unless Bundler.locked_gems.dependencies.key?('ruby-lsp')
```

Did you also confirm that exiting with 1 makes `asyncExec` throw? I don't know what the conditions are for it throwing. Does it throw on any non zero exit status?

andyw8: > Does it throw on any non zero exit status?

Yes: https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback

vinistock: I'd still prefer us to invert the logic though. It feels confusing to me that the error is the true case.

andyw8: Hmmm, to me it feels more natural this way because having `ruby-lsp` in the Gemfile is what we **don't** want.

@st0012 help us break the tie? :grin:

src/client.ts
```diff
@@ -358,12 +358,12 @@ export default class Client implements ClientInterface {
     }
```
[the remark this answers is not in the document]
andyw8: Good point, I've removed it.

PR #515, opened 2023-03-30
Abort existing bundle install if another restart is triggered

### LINKED ISSUE #510: Restarts in the middle of a bundle install don't cancel the one that's already running
If something triggers a restart while installing the gems in the bundle, then we simply open a new bundle install window that basically just hangs until the previous one finishes and then runs bundle install again.

We need to cancel the process that's currently running on restarts.

vinistock:
Closes #510

Use a cancellation source to abort existing bundle install processes if restart is triggered in the middle. This avoids having multiple progress windows (and multiple processes) trying to bundle install.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -30,6 +30,10 @@ export default class Client implements ClientInterface {
   private telemetry: Telemetry;
   private statusItems: StatusItems;
   private outputChannel = vscode.window.createOutputChannel(LSP_NAME);
+  private bundleInstallCancellationSource:
+    | vscode.CancellationTokenSource
+    | undefined;
+
   #context: vscode.ExtensionContext;
   #ruby: Ruby;
   #state: ServerState = ServerState.Starting;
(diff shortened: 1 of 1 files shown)


paracycle (approved):
Amazing! 😍

src/client.ts
```diff
@@ -435,8 +439,23 @@ export default class Client implements ClientInterface {
         title: "Setting up the bundle",
         cancellable: true,
       },
-      (progress, token) =>
-        this.bundleInstall(customGemfilePath, { progress, token })
+      (progress, token) => {
+        if (this.bundleInstallCancellationSource) {
+          this.bundleInstallCancellationSource.cancel();
```
paracycle, line 444: We might want to `.dispose` previous token in this case, otherwise, we might leak memory. So a better call would be:
```suggestion
          this.bundleInstallCancellationSource.dispose(true);
```
Based on the signature of `.dispose` from [here](https://github.com/microsoft/vscode/blob/main/src/vs/base/common/cancellation.ts#L128).

paracycle: Or a sequence of `.cancel()` and `.dispose()` calls, if that API is too new.

PR #520, opened 2023-03-31
Allow specifying a custom bundle gemfile path

vinistock:
Closes Shopify/ruby-lsp#1496

Allow configuring the path to a Gemfile separate from the project to use for the Ruby LSP and other developer tools such as RuboCop. This allows people who are working on projects with older Ruby versions to still use the LSP from a different bundle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,16 @@ The tool to be used for formatting files can be configured with the following se
 "rubyLsp.formatter": "auto"
 ```
 
+#### Using a custom Gemfile
+
+If you are working on a project using an older version of Ruby not supported by Ruby LSP, then you will need you specify a separate `Gemfile` for development tools. You can include the `ruby-lsp` in it and point to that Gemfile by using the following configuration:
+
+**Note**: when using this, gems will not be installed automatically and neither will `ruby-lsp` upgrades.
+
+```jsonc
+"rubyLsp.bundleGemfile": "relative/path/to/Gemfile"
+```
+
 ## Features
 
 ![Ruby LSP demo](extras/ruby_lsp_demo.gif)
--- a/package.json
+++ b/package.json
@@ -164,6 +164,11 @@
             "none"
           ],
           "default": "auto"
+        },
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -70,6 +70,17 @@ The tool to be used for formatting files can be configured with the following se
 "rubyLsp.formatter": "auto"
 ```
 
+#### Using a custom Gemfile
```
andyw8, line 73: Perhaps we can give example of situations where this would be necessary?

andyw8 (approved):

README.md
andyw8: I think re-stating what you said in the PR description is sufficient:

> If you are working on a project using an older version of Ruby not supported by Ruby LSP, then you will need you specify a separate `Gemfile` for development tools. You can include the `ruby-lsp` in it and point to that Gemfile by using the following configuration:

README.md
```diff
@@ -70,6 +70,17 @@ The tool to be used for formatting files can be configured with the following se
...
+
+If you wish to use a Gemfile separate from your project for development tools, you can include the `ruby-lsp` in it and
```
andyw8, line 75:
```suggestion
If you are working on a project using an older version of Ruby not supported by Ruby LSP, then you will need you specify a separate `Gemfile` for development tools. You can include the `ruby-lsp` in it and point to that Gemfile by using the following configuration:
```

README.md
```diff
@@ -70,6 +70,17 @@ The tool to be used for formatting files can be configured with the following se
 "rubyLsp.formatter": "auto"
 ```
 
+#### Using a custom Gemfile
+
+If you are working on a project using an older version of Ruby not supported by Ruby LSP, then you will need you specify a separate `Gemfile` for development tools. You can include the `ruby-lsp` in it and point to that Gemfile by using the following configuration:
+point to that Gemfile by using the following configuration.
```
andyw8, line 76:
```suggestion
```

Issue #514, opened 2023-03-30
Ruby Version Manager: frum


naxels:
Hi team,

Just switched to using Ruby LSP / Ruby extension in VSCode and noticed that
my Version Manager, frum is not supported.

https://github.com/TaKO8Ki/frum

Could support be added, now the extensions fails as it's trying to work with macOS Ruby.

Thank you and keep up the good work!

vinistock:
Hey! Thank you so much for the kind words and the contribution.

I think instead of adding more and more version managers to that list, we should provide a more customizable way for people using less common options. This should also help folks who aren't using a version manager such as in #517.

We'll think about how to support this more widely to allow you to use frum.

PR #522, opened 2023-04-03
Add frum support

naxels:
Hi team,

This resolves https://github.com/Shopify/vscode-ruby-lsp/issues/514

Unfortunately I couldn't exactly discover the need for `this.activate` on frum,
however to be sure added the eval frum init command.

        case VersionManager.Frum:
          await this.activate("eval \"$(frum init)\" && ruby");
          break;

I'm unfamiliar with how to test VSCode extension, so haven't tested the code yet and whether or not the \" escape works.
I did double check to add frum in all places I found another version manager.

Thank you for considering!


naxels:
I just signed the CLA.

Suggestion: add this as a comment/line in the text field when clicking on Create Pull Request.

PR #523, opened 2023-04-03
support rtx version manager

onyxraven:
Support [`rtx`](https://github.com/jdxcode/rtx) as a version manager. 

`rtx` is an asdf-alike, written in rust. It supports and uses the asdf plugin structure, and most of the CLI is compatible with asdf. 

One trick however is that a cheat of linking `asdf` to `rtx` doesn't work because rtx needs the `--` argument separator to be able to pass double-dashed arguments to the exec'd command - which the activate for this plugin uses. So I thought I'd add it as a supported manager.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -73,7 +73,7 @@ export class Ruby {
           await this.activate("rbenv exec ruby");
           break;
         case VersionManager.Rtx:
-          await this.activate("rtx exec ruby --");
+          await this.activate("rtx exec -- ruby");
           break;
         case VersionManager.Rvm:
           await this.activate("rvm-auto-ruby");


vinistock:
Similar to my response here https://github.com/Shopify/vscode-ruby-lsp/issues/514#issuecomment-1494690562, I think we should provide a more customizable option for less common version managers.

andrewthauer:
@Quintasan - The `rtx activate` command is meant to install the shell hook into your shell so that rtx can modify the appropriate path when changing directories. Normally you eval the output of that into your shell on startup which does the chpwd hook (in zsh), etc. Basically it's misleading in the context of the ruby LSP's concept of activate.

You could try just `rtx exec ruby`, but I'm not sure that works. In https://github.com/Shopify/vscode-ruby-lsp/pull/523 I mentioned that I got `rtx` working in a prototype branch, but I needed to do the following for the activation:

> ```ts
> rtx exec ruby@${rubyVersion} -- ruby`
> ```

However, it doesn't look like the `customRubyCommand` supports a version placeholder that can be injected by [readRubyService](https://github.com/Shopify/vscode-ruby-lsp/blob/main/src/ruby.ts#L232). I personally haven't had the time to further try and get the LSP working so I'm not sure, but hopefully this helps a bit.

PR #526, opened 2023-04-03
Source shell configuration files manually

### LINKED ISSUE #505: Failed to activate rbenv environment
Hi, I just trying to use this extension but it fails to activate:

```
Failed to activate rbenv environment: Command failed: /bin/bash -lic 'rbenv exec ruby --disable-gems -rjson -e "printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump(ENV.to_h))"' bash: no se puede establecer el grupo de proceso de terminal (11835): Función ioctl no apropiada para el dispositivo bash: no hay control de trabajos en este intérprete de ordenes No se ha encontrado la orden «rbenv», pero se puede instalar con: sudo apt install rbenv
```
It seems the extension cannot find `rbenv` but I can run the command that the extension is trying to run and is running without errors inside the vscode terminal.

I'm using:
 vscode 1.76.2
 ruby lsp extension 0.2.3
 ruby 2.7.7
 rbenv 1.1.2-40-g62d7798
 ubuntu 20.04

vinistock:
Closes #505

In #349, we tried to switch manually sourcing shell configuration files for using the `-i` flag for interactive mode. Unfortunately, that doesn't seem to always work because interactive mode tries to load other stuff we don't need, such as `~/.profile` or `~/.bash_profile`.

It seems less brittle to just manually source the rc files ourselves.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -125,7 +125,7 @@ export class Ruby {
 
   private async activate(ruby: string) {
     const result = await asyncExec(
-      `${this.shell} -lic '${ruby} --disable-gems -rjson -e "printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump(ENV.to_h))"'`,
+      `${this.shell} -ic '${ruby} --disable-gems -rjson -e "printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump(ENV.to_h))"'`,
       { cwd: this.workingFolder }
     );
 


paracycle (changes requested):
According to the `bash` documentation, the rules of invocation and which files are sourced are defined as follows:

```
INVOCATION
       A login shell is one whose first character of argument zero is a -, or one started with the --login option.

       An interactive shell is one started without non-option arguments (unless -s is specified) and without the -c
       option whose standard input and error are both connected to terminals (as determined by isatty(3)), or one
       started with the -i option.  PS1 is set and $- includes i if bash is interactive, allowing a shell script or a
       startup file to test this state.

       The following paragraphs describe how bash executes its startup files.  If any of the files exist but cannot be
       read, bash reports an error.  Tildes are expanded in filenames as described below under Tilde Expansion in the
       EXPANSION section.

       When bash is invoked as an interactive login shell, or as a non-interactive shell with the --login option, it
       first reads and executes commands from the file /etc/profile, if that file exists.  After reading that file, it
       looks for ~/.bash_profile, ~/.bash_login, and ~/.profile, in that order, and reads and executes commands from the
       first one that exists and is readable.  The --noprofile option may be used when the shell is started to inhibit
       this behavior.

       When an interactive login shell exits, or a non-interactive login shell executes the exit builtin command, bash
       reads and executes commands from the file ~/.bash_logout, if it exists.

       When an interactive shell that is not a login shell is started, bash reads and executes commands from ~/.bashrc,
       if that file exists.  This may be inhibited by using the --norc option.  The --rcfile file option will force bash
       to read and execute commands from file instead of ~/.bashrc.
```

Since our argument zero is not a `-`, if we don't supply the `-l` flag to `bash`, and just supply `-ic`, then it will be an interactive non-login shell and it should not source any of the `profile` files.

I believe other shells should be exhibiting similar behaviour as well.

So, instead of starting to individually handle different shells, I think we should start by changing the flags first.

vinistock:
@paracycle you are correct. I tested on zsh, bash and fish and all of them will source their respective configuration files when using `-ic`. Thank you so much for the precise solution!

paracycle (approved):

PR #539, opened 2023-04-05
Add Ruby debug client

vinistock:
Add a client for the Ruby debug gem. Having the debugger client in the Ruby LSP extension itself allows us to do integrations with the language server quite easily, such as displaying current values in a debug session using [the inline value request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlineValue) or adding a `debug` code lens to launch a test directly in debug mode.

We tested this for launch and attach and it works well. It even allows us to attach to a running process of the Ruby LSP, which makes debugging the server a much much easier task.

**Note**: I have intentionally not added any documentation yet. We need to first test this out and make sure it works properly before documenting and advertising it.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -293,6 +293,8 @@ export default class Client implements ClientInterface {
 
     const gemEntry =
       'gem "ruby-lsp", require: false, group: :development, source: "https://rubygems.org"';
+    const debugEntry =
+      'gem "debug", require: false, group: :development, platforms: :mri, source: "https://rubygems.org"';
 
     // Only try to evaluate the top level Gemfile if there is one. Otherwise, we'll just create our own Gemfile
     if (fs.existsSync(path.join(this.workingFolder, "Gemfile"))) {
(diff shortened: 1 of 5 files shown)


src/debugger.ts
```diff
@@ -0,0 +1,245 @@
...
+  CancellationToken,
+  DebugConfiguration,
+  ProviderResult,
+  workspace,
+  DebugAdapterNamedPipeServer,
+  debug,
+  ExtensionContext,
+  Disposable,
+  window,
+} from "vscode";
```
paracycle, line 21: We've so far been explicitly namespacing classes under the `vscode` namespace, and I think I prefer that since it makes it easier to see what types are coming from VSCode. Should we continue doing the same here?

src/debugger.ts
```diff
@@ -0,0 +1,245 @@
...
+      this.debugProcess = spawn(
+        "bundle",
+        [
+          "exec",
+          "rdbg",
+          "--open",
+          "--command",
+          `--sock-path=${sockPath}`,
+          "--",
+          session.configuration.program,
```
st0012, line 156: Do we want to add a `useBundler` config like `vscode-rdbg` does? When set to true, it prepends `bundle exec` to the program.

src/client.ts
```diff
@@ -307,9 +309,15 @@ export default class Client implements ClientInterface {
         // If it's not a part of the bundle, add it to the custom Gemfile
         gemfile.push(gemEntry);
       }
+
+      // If debug is not in the bundle, add it to allow debugging
+      if (!(await this.debugIsInBundle())) {
```
st0012, line 314: 👍

src/debugger.ts
vinistock: Yeah, agreed.

st0012 (commented):
I think this also closes https://github.com/Shopify/ruby-dev-exp-issues/issues/745?

src/debugger.ts
vinistock: I don't think it's necessary. In the configurations, you can just use
```json
{
  "program": "bundle exec ruby ${file}"
}
```
I think it's simpler than introducing another field just to concatenate `bundle exec` for the user.

st0012: Yeah sure let's keep it simple first.

vinistock:
I've made two improvements: added `platforms: :mri` to the debug gem entry in the Gemfile to avoid issues for other Ruby implementations and merged the implementations to check if a gem is a dependency into a single function.

st0012 (approved):

PR #544, opened 2023-04-06
Use valid paths for Ruby test

vinistock:
I dig deeper to figure out why CI keeps failing on #507. The reason is because now that we try to execute Ruby, we're doing so with a fake system path, which we pass to `asyncExec` and then fails. Unfortunately the test output doesn't hint at that, so it wasn't clear what the issue was.

We can only use paths that exist in this test. And we unfortunately can't create paths because the file system is read only while running tests. So this PR uses the real process PWD to run the main test and simply removes the alternative one.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -7,7 +7,7 @@ import { Ruby } from "./ruby";
 let client: Client;
 
 export async function activate(context: vscode.ExtensionContext) {
-  const ruby = new Ruby();
+  const ruby = new Ruby(context);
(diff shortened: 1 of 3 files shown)


andyw8 (approved):

st0012:
> We can only use paths that exist in this test. And we unfortunately can't create paths because the file system is read only while running tests. 

Is it a limitation on GH actions? I though we can still create files or folders on it, at least under certain paths with `Dir.mktmpdir`.

PR #546, opened 2023-04-08
Make the extension work on Windows, depends on ruby-lsp PR#620

janbiedermann:


src/client.ts
```diff
@@ -54,6 +56,9 @@ export default class Client implements ClientInterface {
     vscode.window.onDidCloseTerminal((terminal: vscode.Terminal): void => {
       if (terminal === this.terminal) this.terminal = undefined;
     });
+    if (os.platform().indexOf('win') == 0) {
```
vinistock, line 59: My understanding is that it will return `win32` (even on 64 bit). Can we use equality here instead?
```typescript
os.platform() === "win32"
```

src/client.ts
```diff
@@ -35,6 +36,7 @@ export default class Client implements ClientInterface {
     | undefined;
 
   private terminal: vscode.Terminal | undefined;
+  private windows = false;
```
vinistock, line 39: Let's avoid this extra property and just check every time we restart the server.

src/ruby.ts
```diff
@@ -35,6 +39,12 @@ export class Ruby {
   ) {
     this.context = context;
     this.workingFolder = workingFolder;
+    if (os.platform().indexOf('win') == 0) {
```
vinistock, line 42: Same comment here.

src/ruby.ts
```diff
@@ -35,6 +39,12 @@ export class Ruby {
...
+      this.shell = "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
```
vinistock, line 43: Why do we need to set the shell manually? Does Windows not set this in any way?

src/ruby.ts
```diff
@@ -136,8 +146,9 @@ export class Ruby {
   }
 
   private async activate(ruby: string) {
-    const result = await asyncExec(
-      `${this.shell} -ic '${ruby} --disable-gems -rjson -e "printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump(ENV.to_h))"'`,
+    const result = await asyncExec(this.windows ?
+      `${this.shell} ${this.shell_opt_activate}  ${ruby} --disable-gems -rjson -e 'printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump(ENV.to_h))' ` :
+      `${this.shell} ${this.shell_opt_activate} '${ruby} --disable-gems -rjson -e "printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump(ENV.to_h))"'`,
```
vinistock, line 151: What's the difference between the two here? Just the single quotes?

Why can't we use quotes on powershell?

src/ruby.ts
```diff
@@ -284,7 +295,7 @@ export class Ruby {
 
   private async toolExists(tool: string) {
     try {
-      await asyncExec(`${this.shell} -lic '${tool} --version'`);
+      await asyncExec(`${this.shell} ${this.shell_opt_exist} '${tool} --version'`);
```
vinistock, line 298: Here, we actually forgot to remove the `-l` flag. It should also just be `-ic` similar to the Ruby activation command.

src/client.ts
```diff
@@ -92,13 +97,13 @@ export default class Client implements ClientInterface {
     };
 
     const executable: Executable = {
-      command: "bundle",
+      command: this.windows ? "bundle.bat" : "bundle",
```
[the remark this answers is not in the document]
janbiedermann: ChildProcess.spawn doesn't use a shell

src/ruby.ts
janbiedermann: Right, i agree, though then ChildProcess.exec will use cmd.exe, which has problems with " (double quotes) (can't really escape them). Thats why i use powershell here.

src/client.ts
janbiedermann: sure

src/ruby.ts
janbiedermann: I use powershell here, because otherwise node would use cmd.exe which makes problems with the string, can't escape " (double quotes) in cmd.exe. So wither the cmd string breaks, or the ruby string.

src/ruby.ts
janbiedermann: So the problem is not powershell, the problem is cmd.exe. I use poweshell to avoid cmd.exe problems.

PR #548, opened 2023-04-10
Add bug and feature issue templates

vinistock:
Add a bug report and feature proposal issue templates. I used the new form features from GitHub to make it easier to include relevant information.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_template.yml
+++ b/.github/ISSUE_TEMPLATE/feature_template.yml
@@ -1,0 +1,40 @@
+---
+name: Ruby LSP feature proposal
+description: Propose a new feature for the Ruby LSP
+labels:
+  - enhancement
+body:
+  - type: markdown
+    attributes:
+      value: >
+        Thank you for proposing a new feature! Please answer the following questions to help us the proposal.
+  - type: checkboxes
+    attributes:
+      label: This feature is VS Code specific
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/feature_template.yml
```diff
@@ -0,0 +1,47 @@
...
+      value: >
+        Thank you for proposing a new feature! Please answer the following questions to help us the proposal.
+  - type: checkboxes
+    attributes:
+      label: This feature is VS Code specific
+      description: Non VS Code specific features should be proposed to the [Ruby LSP server](https://github.com/Shopify/ruby-lsp/issues/new)
+      options:
+        - label: VS Code specific
+          required: true
+  - type: checkboxes
```
andyw8, line 18: I'd maybe drop this one to keep the number of form fields to a reasonable length.

st0012: Yeah I agree. Perhaps let's not have it yet and see if it's really needed?

st0012 (approved):
In bug report template, it may be helpful to have a screenshot on the status panel too? It'll provide a bunch useful information without users typing them.
But then I realised that it disappears when I hit `shift`, which means the screenshotting shortcut on mac doesn't work with it. Any thoughts on this?

PR #549, opened 2023-04-10
Add formatter status item

andyw8:
This PR adds a new status item to indicate to indicate which formatter is in use when `"rubyLsp.formatter": "auto"` is set.

It does this by checking for a **direct** dependency on RuboCop, or on Syntax Tree.

To support RuboCop-based gems such as `rubocop-shopify`, it checks for any gems beginning `rubocop-`.

The behaviour is intend to mirror the client-side behaviour being implemented in https://github.com/Shopify/ruby-lsp/pull/654. This essentially means we have the same behaviour duplicated across the gem and the extension. (We could not find a good way to have the behaviour only in a single place).

For consistency, we should aim to launch this ~at the same time as~ after [#614](https://github.com/Shopify/ruby-lsp/pull/654).

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,13 @@ vscode.commands.registerCommand(
 );
 ```
 
+## Formatting
+
+When `rubyLsp.formatter` is set to `auto`, Ruby LSP tries to determine which formatter to use.
+
+If the bundle has a **direct** dependency on a supported formatter, such as `rubocop` or `syntax_tree`, that will be used.
+Otherwise, formatting will be disabled and you will need add one to the bundle.
+
 ## Contributing
 
 Bug reports and pull requests are welcome on GitHub at https://github.com/Shopify/vscode-ruby-lsp.
--- a/src/client.ts
+++ b/src/client.ts
@@ -40,6 +40,7 @@ export default class Client implements ClientInterface {
   #context: vscode.ExtensionContext;
   #ruby: Ruby;
   #state: ServerState = ServerState.Starting;
+  #formatter: string;
 
   constructor(
     context: vscode.ExtensionContext,
@@ -52,6 +53,7 @@ export default class Client implements ClientInterface {
     this.testController = testController;
     this.#context = context;
     this.#ruby = ruby;
+    this.#formatter = "";
     this.statusItems = new StatusItems(this);
     this.registerCommands();
     this.registerAutoRestarts();
@@ -207,6 +209,7 @@ export default class Client implements ClientInterface {
     );
 
     await this.client.start();
+    await this.determineFormatter();
 
     this.state = ServerState.Running;
   }
@@ -259,6 +262,27 @@ export default class Client implements ClientInterface {
 
   private set ruby(ruby: Ruby) {
     this.#ruby = ruby;
+  }
+
+  get formatter(): string {
+    return this.#formatter;
+  }
+
+  async determineFormatter() {
+    const configuration = vscode.workspace.getConfiguration("rubyLsp");
+    const configuredFormatter: string = configuration.get("formatter")!;
+
+    if (configuredFormatter === "auto") {
+      if (await this.projectHasDependency(/^rubocop/)) {
+        this.#formatter = "rubocop";
+      } else if (await this.projectHasDependency(/^syntax_tree$/)) {
+        this.#formatter = "syntax_tree";
+      } else {
+        this.#formatter = "none";
+      }
+    } else {
+      this.#formatter = configuredFormatter;
+    }
   }
 
   get context(): vscode.ExtensionContext {
@@ -341,7 +365,7 @@ export default class Client implements ClientInterface {
       gemfile.push('eval_gemfile(File.expand_path("../Gemfile", __dir__))');
 
       // If the `ruby-lsp` exists in the bundle, add it to the custom Gemfile commented out
-      if (await this.projectHasDependency("ruby-lsp")) {
+      if (await this.projectHasDependency(/^ruby-lsp$/)) {
         // If it is already in the bundle, add the gem commented out to avoid conflicts
         gemfile.push(`# ${gemEntry}`);
       } else {
@@ -350,7 +374,7 @@ export default class Client implements ClientInterface {
       }
 
       // If debug is not in the bundle, add it to allow debugging
-      if (!(await this.projectHasDependency("debug"))) {
+      if (!(await this.projectHasDependency(/^debug$/))) {
         gemfile.push(debugEntry);
       }
     } else {
@@ -424,7 +448,7 @@ export default class Client implements ClientInterface {
     return Object.keys(features).filter((key) => features[key]);
   }
 
-  private async projectHasDependency(gemName: string): Promise<boolean> {
+  private async projectHasDependency(gemName: RegExp): Promise<boolean> {
     try {
       // We can't include `BUNDLE_GEMFILE` here, because we want to check if the project's bundle includes the
       // dependency and not our custom bundle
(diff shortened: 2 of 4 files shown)


src/client.ts
```diff
@@ -235,6 +239,32 @@ export default class Client implements ClientInterface {
...
 
+  async determineFormatter() {
+    const configuration = vscode.workspace.getConfiguration("rubyLsp");
+    const configuredFormatter: string = configuration.get("formatter")!;
+    switch (configuredFormatter) {
+      case "auto":
+        if (await this.projectHasDependency("rubocop")) {
+          this.client!.clientOptions.initializationOptions.formatter =
+            "rubocop";
+          this.client!.restart();
```
andyw8, line 250: @vinistock I'm wondering if there's a better way to handle this. Since the `determineFormatter` run asynchronously after the initialization, I need to load the updated config by restarting.

An alternative might be to have `determineFormatter` run synchronously but that would delay the initial startup time.

README.md
```diff
@@ -189,6 +189,13 @@ vscode.commands.registerCommand(
 );
 ```
 
+## Formatting
+
+When `rubyLsp.formatter` is set to `auto`, the extension will try to determine which formatter to use.
+
+If the bundle has a **direct** dependency on `rubocop` or `syntax_tree`, that will be used.
+Otherwise, formatting will be disabled and you will need explicity choose one.
```
vinistock, line 197:
```suggestion
Otherwise, formatting will be disabled and you will need add one to the bundle.
```

README.md
```diff
@@ -189,6 +189,13 @@ vscode.commands.registerCommand(
 );
 ```
 
+## Formatting
+
+When `rubyLsp.formatter` is set to `auto`, the extension will try to determine which formatter to use.
+
+If the bundle has a **direct** dependency on `rubocop` or `syntax_tree`, that will be used.
```
vinistock, line 196:
```suggestion
If the bundle has a **direct** dependency on a formatter, such as `rubocop` or `syntax_tree`, that will be used.
```

src/client.ts
vinistock: Sorry, I don't understand why you need to restart. We already have a restart watcher if any Ruby LSP related configurations are changed. Is it not restarting when you change the config?

src/client.ts
```diff
@@ -235,6 +239,32 @@ export default class Client implements ClientInterface {
...
+          return "rubocop";
+        } else if (await this.projectHasDependency("syntax_tree")) {
+          this.client!.clientOptions.initializationOptions.formatter =
```
vinistock, line 253: This is not necessary. We're already setting the initialization option in line 119. Whatever is configured, we send to the server. The server will have to decide which formatter to use.

The code on the client should just decide which information to display on the status item, not change what we send the server.

andyw8: Got it. I was mistakenly thinking that this would act the 'source of truth' to determine what formatter to use, but instead we can just the gem mirror this behaviour.

src/client.ts
```diff
@@ -235,6 +239,24 @@ export default class Client implements ClientInterface {
     this.#ruby = ruby;
   }
 
+  async determineFormatter() {
+    const configuration = vscode.workspace.getConfiguration("rubyLsp");
+    const configuredFormatter: string = configuration.get("formatter")!;
+    switch (configuredFormatter) {
+      case "auto":
+        if (await this.projectHasDependency(/^rubocop?-/)) {
```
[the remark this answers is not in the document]
vinistock: I think that's a safe assumption. However, I think the regex needs to be `/^rubocop-?/` otherwise `rubocop` doesn't match.

st0012 (commented):
When I opened `ruby-lsp` with this branch, it choses `rubocop` by default.
Perhaps we'd want to favour `syntax-tree` when both are present in the Gemfile?

Screenshot 2023-04-24 at 17 09 46

Also, when opening `IRB`, it still shows the formatter is `rubocop`, even though it doesn't have it as dependency at all 🤔

andyw8:
> When I opened ruby-lsp with this branch, it choses rubocop by default.

Yeah that's because `syntax_tree` is not considered a direct dependency, it's a dependency of `ruby-lsp`. I'm not sure there's a good around that.

> Also, when opening IRB, it still shows the formatter is rubocop, even though it doesn't have it as dependency at all 

Odd, I'll look into that.

andyw8:
I've pushed a fix.

st0012:
It still doesn't work as expected. In below's cases, I've both tried 1) restart LSP and 2) reload window

### In `ruby-lsp`

I specifically configured `rubyLsp.formatter` to `none`, and:
- It does stop formatting
- But the status doesn't show `none`

Screenshot 2023-04-24 at 18 16 04


### In IRB

Screenshot 2023-04-24 at 18 12 18

It shows that the formatter is `none`, but it still formats my code.

src/client.ts
```diff
@@ -238,6 +243,21 @@ export default class Client implements ClientInterface {
...
+    if (configuredFormatter === "auto") {
+      if (await this.projectHasDependency(/^rubocop-?/)) {
+        this.formatter = "rubocop";
+      } else if (await this.projectHasDependency("syntax_tree")) {
+        this.formatter = "syntax_tree";
+      } else {
+        this.formatter = "none";
+      }
+    }
+  }
```
st0012, line 259: After pairing with @andyw8, a change like this is needed to display the formatter when the configured value isn't `"auto"`

```suggestion
  async determineFormatter() {
    const configuration = vscode.workspace.getConfiguration("rubyLsp");
    const configuredFormatter: string = configuration.get("formatter")!;

    if (configuredFormatter === "auto") {
      if (await this.projectHasDependency(/^rubocop-?/)) {
        this.formatter = "rubocop";
      } else if (await this.projectHasDependency("syntax_tree")) {
        this.formatter = "syntax_tree";
      } else {
        this.formatter = "none";
      }
    } else if (configuredFormatter !== undefined) {
      this.formatter = configuredFormatter;
    }
  }
```

st0012 (approved):

src/client.ts
```diff
@@ -397,15 +419,23 @@ export default class Client implements ClientInterface {
...
-  private async projectHasDependency(gemName: string): Promise<boolean> {
+  private async projectHasDependency(
+    gemName: string | RegExp
+  ): Promise<boolean> {
+    let command = null;
+    if (typeof gemName === "string") {
+      command = `Bundler.locked_gems.dependencies.key?('${gemName}')`;
+    } else {
+      command = `Bundler.locked_gems.dependencies.keys.grep(${gemName}).any?`;
+    }
```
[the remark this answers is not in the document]
andyw8: Yeah, makes sense. I'll update the gem PR to match.

src/client.ts
```diff
@@ -397,15 +421,17 @@ export default class Client implements ClientInterface {
     return Object.keys(features).filter((key) => features[key]);
   }
 
-  private async projectHasDependency(gemName: string): Promise<boolean> {
+  private async projectHasDependency(gemName: RegExp): Promise<boolean> {
+    let command = null;
+    command = `Bundler.locked_gems.dependencies.keys.grep(${gemName}).any?`;
```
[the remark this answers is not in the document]
andyw8: Makes sense, done.

src/client.ts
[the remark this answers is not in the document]
pl-at-shopify: Actually, we can simplify it even more to `/^rubocop/` as `/^rubocop-?/` is matching the same set of string as there is no trailing `$` in the end and the `-?` does not restrict the set much, if I understand it correctly.

vinistock:
Instead of attaching functionality to a specific server version, I would rather we just show what is configured then (not what was inferred based on the `auto` option).

It would match what we do for the version manager and remove some complexity in this PR as well (since you no longer will need the client interface to include formatter).

andyw8:
> Instead of attaching functionality to a specific server version, I would rather we just show what is configured then (not what was inferred based on the auto option).

🤔 I was hoping that the status item would help give insight into what the gem was doing. But I can change it to this.

andyw8:
See https://github.com/Shopify/vscode-ruby-lsp/pull/586 for the simplified version.

vinistock:
> 🤔 I was hoping that the status item would help give insight into what the gem was doing. But I can change it to this.

That's a good point indeed. Given that we upgrade the gem automatically and people will eventually catch up to the right version, do you think sometimes displaying incorrect information for old server versions is a big issue?

We can also release the server first to change the behaviour and get people upgrading and then release the status item later.

src/client.ts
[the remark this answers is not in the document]
andyw8: Thanks, I've updated to use that.

vinistock:
Yeah, looks good 🚢

PR #559, opened 2023-04-11
Remove none from bug template

vinistock:
### Motivation

Apparently it's a reserved word.


st0012:
It still errors: `Checkboxes must have unique labels.`

I recommend viewing the file in the branch, which will also give you those hints: https://github.com/Shopify/vscode-ruby-lsp/blob/vs/remove_none_from_template/.github/ISSUE_TEMPLATE/bug_template.yml

vinistock:
Thanks, that was helpful.

PR #562, opened 2023-04-14
Add test controller implementation

adisonlampert:
### Motivation

Closes #498 
Closes #382

Screenshot 2023-04-14 at 2 07 41 PM

A continuation of #506 to use the VS Code Test Explorer 

Update: After talking with Vini, we decided not to implement opening the terminal on run and outputting the test results.

### Implementation

- Created a new class `TestController`
- Added middleware to create a VS Code [`TestItem`](https://code.visualstudio.com/api/references/vscode-api#TestItem) for each test in an open file.
- The test items are added to a [`TestController`](https://code.visualstudio.com/api/references/vscode-api#TestController) which is the entry point to discover and execute tests.
- When you either run the test from the gutter of the file editor tab or from the Test Explorer tab, executes the method `runHandler` using the [`TestRunProfile`](https://code.visualstudio.com/api/references/vscode-api#TestRunProfile).
- Updated the method registered as `rubyLsp.runTest` to `runOnClick` which finds the `TestItem` associated with the name of the test, opens the Test Explorer to display that item, and runs the test using the `runHandler` method.

### Manual Tests

1. Debug this branch on `ruby-lsp`
2. Open up a test file (ex. `test/document_test.rb`)
3. Observe there is a "Run" button next to the class definition and test method definitions
4. Open the Test Explorer tab, observe the class and test methods are runnable there
5. Click the "Run" button next to a test method; obverse that it passes
7. Click the "Run" button in the Test Explorer for the class; observe that all the tests are run and pass
8. Change one of the tests so that it will fail
9. Click the "Run" text above the method definition; obverse that it fails and that a failure message appears

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -36,7 +36,6 @@ export default class Client implements ClientInterface {
     | vscode.CancellationTokenSource
     | undefined;
 
-  private terminal: vscode.Terminal | undefined;
   private testController: TestController;
   #context: vscode.ExtensionContext;
   #ruby: Ruby;
@@ -45,23 +44,17 @@ export default class Client implements ClientInterface {
   constructor(
     context: vscode.ExtensionContext,
     telemetry: Telemetry,
-    ruby: Ruby
+    ruby: Ruby,
+    testController: TestController
   ) {
     this.workingFolder = vscode.workspace.workspaceFolders![0].uri.fsPath;
     this.telemetry = telemetry;
+    this.testController = testController;
     this.#context = context;
     this.#ruby = ruby;
     this.statusItems = new StatusItems(this);
     this.registerCommands();
     this.registerAutoRestarts();
-    this.testController = new TestController(
-      context,
-      this.workingFolder,
-      this.ruby
-    );
-    vscode.window.onDidCloseTerminal((terminal: vscode.Terminal): void => {
-      if (terminal === this.terminal) this.terminal = undefined;
-    });
   }
 
   async start() {
@@ -126,16 +119,21 @@ export default class Client implements ClientInterface {
       },
       middleware: {
         provideCodeLenses: async (document, token, next) => {
-          if (this.client) {
-            const response = await next(document, token);
-
-            if (response) {
-              this.testController.createTestItems(response as CodeLens[]);
-            }
-
-            return response;
+          if (!this.client) {
+            return null;
           }
-          return null;
+
+          const response = await next(document, token);
+
+          if (response) {
+            this.testController.createTestItems(
+              response.filter(
+                (codeLens) => (codeLens as CodeLens).data.type === "test"
+              ) as CodeLens[]
+            );
+          }
+
+          return response;
         },
         provideOnTypeFormattingEdits: async (
           document,
@@ -251,7 +249,6 @@ export default class Client implements ClientInterface {
   }
 
   dispose() {
-    this.testController.dispose();
     this.client?.dispose();
     this.outputChannel.dispose();
   }
@@ -290,21 +287,11 @@ export default class Client implements ClientInterface {
         Command.Update,
         this.updateServer.bind(this)
       ),
-      vscode.commands.registerCommand(Command.RunTest, this.runTest.bind(this)),
       vscode.commands.registerCommand(
         Command.DebugTest,
         this.debugTest.bind(this)
       )
     );
-  }
-
-  private runTest(_path: string, _name: string, command: string) {
-    if (this.terminal === undefined) {
-      this.terminal = vscode.window.createTerminal({ name: "Run test" });
-    }
-
-    this.terminal.show();
-    this.terminal.sendText(command);
   }
 
   private debugTest(_path: string, _name: string, command: string) {
--- a/src/client.ts
+++ b/src/client.ts
@@ -10,11 +10,13 @@ import {
   LanguageClient,
   Executable,
   RevealOutputChannelOn,
+  CodeLens,
 } from "vscode-languageclient/node";
 
 import { Telemetry } from "./telemetry";
 import { Ruby } from "./ruby";
 import { StatusItems, Command, ServerState, ClientInterface } from "./status";
+import { TestController } from "./testController";
 
 const LSP_NAME = "Ruby LSP";
 const asyncExec = promisify(exec);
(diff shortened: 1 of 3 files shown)


vinistock (approved):
This looks amazing.

Is there a way for us to show output for tests that pass? Imagine you're trying to debug something and go
```ruby
def test_something
  # Can we see the output somewhere in the test explorer?
  puts variable_im_trying_to_understand
end
```

vinistock:
In fact, I think there's a panel for test results. We might be able to reveal that panel automatically to show output.

adisonlampert:
I updated `runHandler` so that it only runs test methods which solved the issue I described above. 

In the VS Code Testing API Guide, [the recommendation](https://code.visualstudio.com/api/extension-guides/testing#:~:text=The%20TestItem%20interface%20is%20simple%20and%20doesn%27t%20have%20room%20for%20custom%20data.%20If%20you%20need%20to%20associate%20extra%20information%20with%20a%20TestItem%2C%20you%20can%20use%20a%20WeakMap%3A) is to use `WeakMap` to add additional information to `TestItem` objects. Previously, we were using the `TestItem` description to store the test command (left version). I implemented `WeakMap` (right version). I feel like the right version is better because you can see the time it took for the test to run after it has finished and looks cleaner + I feel that it is more in line with the intended use of `TestItem`.

<p float="left">
Screenshot 2023-04-18 at 12 07 23 PM
Screenshot 2023-04-18 at 12 05 02 PM
</p>

vinistock:
It seems that every time a test runs, a new test output tab is created. It is possible to reuse the same one?

src/testController.ts
```diff
@@ -0,0 +1,207 @@
...
+        }
+      }
+    });
+  }
+
+  async runHandler(
+    request: vscode.TestRunRequest,
+    token: vscode.CancellationToken
+  ) {
+    const run = this.testController.createTestRun(request, undefined, true);
```
adisonlampert, line 100: @vinistock 

The same output terminal was being used for tests run through the Test Explorer UI, but a new terminal was created each time a test was run through code lens (i.e., through `rubyLsp.runTest`). 

Since I am using `run.appendOutput` to log the test results, I knew that `run` must contain some reference to a terminal and that terminal was the same for anything run through the Test Explorer UI but not for code lens runs. 

By debugging, I observed that every Text Explorer `TestRun` had `persisted?` set to `true` even though we were setting it to false on line 100. However, every code lens `TestRun` had `persisted?` set to `false`. This was the only difference I could find between the two runs so I set `persisted?` to true here and that fixed the issue of opening multiple terminals.

src/testController.ts
```diff
@@ -0,0 +1,207 @@
...
+    this.testRunProfile = this.testController.createRunProfile(
+      "Run",
+      vscode.TestRunProfileKind.Run,
+      (request, token) => {
+        this.runHandler(request, token);
+      },
+      true
+    );
+
+    vscode.commands.executeCommand("testing.clearTestResults");
```
adisonlampert, line 43: I added in this line to clear the test results each time the LSP is started/restarted.

BTW, `testing.clearTestResults` and `testing.showMostRecentOutput` are not in the VS Code API documentation (or I couldn't find them in there). I got these commands from the source code [here](https://github.com/microsoft/vscode/blob/8f74fbfd1f2d8f6268a42df131726b218aafe511/src/vs/workbench/contrib/testing/common/constants.ts#L53). There are other test-related commands that might be useful later so I wanted to document it.

src/testController.ts
```diff
@@ -0,0 +1,197 @@
...
+          this.runOnClick(name);
+        }
+      )
+    );
+  }
+
+  createTestItems(response: CodeLens[]) {
+    this.testController.items.forEach((test) => {
+      this.testController.items.delete(test.id);
+    });
```
vinistock, line 59: Should we also clear the `this.testCommands` weakmap here?

vinistock (approved):

src/client.ts
```diff
@@ -117,6 +125,18 @@ export default class Client implements ClientInterface {
...
+          if (this.client) {
+            const response = await next(document, token);
+
+            if (response) {
+              this.testController.createTestItems(response as CodeLens[]);
+            }
+
+            return response;
+          }
+          return null;
```
paracycle, line 138: Nitpick, can we invert the check to reduce nesting here?
```suggestion
          if (!this.client) {
            return null;
          }
          
          const response = await next(document, token);

          if (response) {
            this.testController.createTestItems(response as CodeLens[]);
          }

          return response;
```

src/testController.ts
```diff
@@ -0,0 +1,197 @@
...
+    this.ruby = ruby;
+
+    this.testController = vscode.tests.createTestController(
+      "rubyTests",
+      "Ruby Tests"
+    );
+
+    this.testCommands = new WeakMap<vscode.TestItem, string>();
+
+    this.testRunProfile = this.testController.createRunProfile(
```
paracycle, line 34: Can we have more than 1 profile defined? Could we also have a run profile that enables debugging? (Probably not as part of this PR, though)

src/client.ts
```diff
@@ -51,6 +54,11 @@ export default class Client implements ClientInterface {
     this.statusItems = new StatusItems(this);
     this.registerCommands();
     this.registerAutoRestarts();
+    this.testController = new TestController(
+      context,
+      this.workingFolder,
+      this.ruby
+    );
```
paracycle, line 61: I don't think the `TestController` creation should be the responsibility of the `Client` class, which is supposed to be just the LSP client. I think we can move the creation of the test controller to `extension.ts` and inject it to the client as needed.

src/client.ts
```diff
@@ -117,6 +125,18 @@ export default class Client implements ClientInterface {
         formatter: configuration.get("formatter"),
       },
       middleware: {
+        provideCodeLenses: async (document, token, next) => {
+          if (this.client) {
+            const response = await next(document, token);
+
+            if (response) {
+              this.testController.createTestItems(response as CodeLens[]);
```
paracycle, line 133: Is this assuming that all code lenses will be testing related? What happens if the LSP server decides to add other kinds of code lenses? Will this code break?

src/testController.ts
vinistock: Yeah, we intend to create a debugging profile to launch debugging sessions from tests.

vinistock (approved):
🚀

danielpgross:
This is really exciting, great work! 🎉

PR #569, opened 2023-04-19
The extension should warn when debuggee process doesn't exit with code 0

st0012:
The current logic doesn't show the error window when the debuggee process exists with code 1.

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -192,7 +192,7 @@ export class Debugger
       // code zero here is because debug actually exits with 1 if the user cancels the debug session, which is not
       // actually an error
       this.debugProcess.on("exit", (code) => {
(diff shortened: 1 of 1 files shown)


andyw8 (approved):

vinistock (changes requested):
When someone stops the debugging session manually, by clicking the stop button, the debugger always exits with status 1 (as the comment above this code mentions).

If we want to print something, let's at least not say that status 1 is an error, given that the user manually stopped the process

vinistock:
I think the problem is that the debug gem shouldn't exit with code `1` if the user is intentionally disconnecting. Not sure if there's a reason behind the decision, but I think it should just exit with `0`.

PR #570, opened 2023-04-19
Fix debug client launch

vinistock:
### Motivation

There were a few issues with the debugger client that are fixed in this PR.

### Implementation

1. The `provideDebugConfigurations` method is used to create a `launch.json` file. We shouldn't inject the Ruby environment there since it shouldn't be hard coded
2. We were missing the `wait for debugger connection` message. What happened is that when that message came in, we concatenated it into `initialMessage`, but then never resolve the promise until we received another message in `stderr`, which doesn't always happen. I changed to use an `initialized` flag so that we know when to concatenate
3. We were not adding 1 to the socket file name
4. We were incorrectly returning early from `deactivate` and not disposing of the debug client properly

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -64,7 +64,6 @@ export class Debugger
         request: "launch",
         // eslint-disable-next-line no-template-curly-in-string
         program: "ruby ${file}",
-        env: this.ruby.env,
       },
       {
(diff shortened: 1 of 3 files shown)


st0012 (approved):
It's working for me now 👍

PR #571, opened 2023-04-21
Use general middleware to report telemetry

### LINKED ISSUE #310: Explore end-to-end performance telemetry
Currently, we receive request benchmarks from the server gem. Unfortunately, that information is incomplete, since it only includes the amount of time it took for the server to process the request, which will not reflect things like IO and backlogging of the request queue.

Explore if we can add end-to-end telemetry that completely measures the time it took from sending the request to receiving a response.

Note: the easiest way to implement this would be to have a middleware for all requests that sends telemetry data - however, this is not currently supported (we only have request specific middleware). There is a [suggestion to add this to the language server node package](https://github.com/microsoft/vscode-languageserver-node/issues/918) tagged as help wanted. Maybe we can take a stab at implementing it so that any LSP can benefit from it?

vinistock:
### Motivation

Closes #310

Related to https://github.com/Shopify/ruby-lsp/pull/650.

Use the new general middleware to benchmark requests and report telemetry entirely from the client side. This allows us to remove telemetry events from the server and cut the amount of IO it does in half.

### Implementation

Added the new middleware and used `perf_hooks` to measure the performance of executing the requests. If a request fails, the server returns extra information about the errors in the response which we use to report.

Added a function to recursively sanitize URIs that are inside parameters so that it doesn't include the `HOME` part of the directory.

Added a function to figure out the version of the `ruby-lsp` gem.

I'd love to get some feedback on these three points. I'm especially not crazy about how we're figuring out the server version.

**Note**: we may want to hold back until there's a stable release of the languageclient package.

### Manual Tests

1. Start the LSP in development mode on this branch and using https://github.com/Shopify/ruby-lsp/pull/650
2. Play around with the LSP
3. Look at the debug console and verify that events are being printed properly

### DIFF
--- a/package.json
+++ b/package.json
@@ -351,6 +351,6 @@
     "typescript": "^5.0.4"
   },
   "dependencies": {
-    "vscode-languageclient": "^8.1.0"
+    "vscode-languageclient": "^8.2.0-next.0"
   }
 }
--- a/src/client.ts
+++ b/src/client.ts
@@ -3,6 +3,8 @@ import fs from "fs";
 import readline from "readline";
 import { promisify } from "util";
 import { exec, spawn } from "child_process";
+import { performance as Perf } from "perf_hooks";
+import { randomUUID } from "crypto";
 
 import * as vscode from "vscode";
 import {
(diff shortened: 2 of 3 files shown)


andyw8:
Moving to backlog as discussed at team meeting on 2023-06-12.

PR #583, opened 2023-04-25
Add custom Ruby activation

vinistock:
### Motivation

Closes Shopify/ruby-lsp#1571
Closes Shopify/ruby-lsp#1577

We need to provide a way for custom Ruby activation for less common version managers or installers that don't activate the Ruby environment, such as the ones on Windows.

This PR adds a new configuration that can be used to define the custom activation and also fixes the bug in Shopify/ruby-lsp#1577. Testing these changes surfaced the issue in my editor and I managed to find what I believe is the root cause.

### Implementation

- Added a new section in the README to explain the custom activation
- Added a new configuration for the custom activation
- In `Ruby`, added a new version manager option and start executing the configured custom command before invoking ruby

Finally, I changed the way `projectHasDependency` is implemented a bit. There are two fixes included
1. We shouldn't use `BUNDLE_GEMFILE` when checking if a project has a dependency. We're interested in knowing if the main bundle has the dependency - not our custom bundle. So we can't include `BUNDLE_GEMFILE` or else we're basically asking if our own `Gemfile` includes the dependency
2. For some reason, using `bundle exec ruby` always seems to load an incorrect bundle. Removing `bundle exec` and just requiring bundler makes the check work properly

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ by clicking `Change version manager` in the language status center or by changin
 // Available options are
 // "auto" (select version manager automatically)
 // "none" (do not use a version manager)
+// "custom" (use rubyLsp.customRubyCommand for finding/activating Ruby)
 // "asdf"
 // "chruby"
 // "rbenv"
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -399,16 +399,18 @@ export default class Client implements ClientInterface {
 
   private async projectHasDependency(gemName: string): Promise<boolean> {
     try {
+      // We can't include `BUNDLE_GEMFILE` here, because we want to check if the project's bundle includes the
+      // dependency and not our custom bundle
+      const withoutBundleGemfileEnv = { ...this.ruby.env };
```
paracycle, line 404: The idiomatic way to do this is by using [Rest in Object Destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Rest_in_Object_Destructuring) operation:
```suggestion
      const { BUNDLE_GEMFILE, ...withoutBundleGemfileEnv } = this.ruby.env;
```

vinistock: Very nice.

README.md
```diff
@@ -58,6 +59,28 @@ To make sure that the Ruby LSP can find the version manager scripts, make sure t
...
+```jsonc
+{
+  // Don't forget to set the manager to custom when using this option
+  "rubyLsp.rubyVersionManager": "custom",
+
+  // Using a different version manager than the ones included by default
+  "rubyLsp.customRubyCommand": "my_custom_version_manager activate",
+
+  // Adding a custom Ruby bin folder to the PATH
+  "rubyLsp.customRubyCommand": "PATH=/path/to/ruby/bin:$PATH"
```
andyw8, line 80: btw on Windows this may need `;` instead of `:`.

vinistock: Yeah, users will need to write their own activation based on their machines.

dirceu (approved):

bitwise-aiden (approved):

PR #584, opened 2023-04-26
Support debugging tests with code lens

st0012:
### Motivation

We want to let users debug their tests with code lens with well.

### Implementation

With https://github.com/Shopify/ruby-lsp/pull/657, we can display a `Debug` code lens on top of test classes/methods. Once clicked, the tests would be run with debugger attached.

### Manual Tests

1. Click `Run Extension` in the debug panel
2. Make sure the `Extension Development Host` runs `ruby-lsp` and is using the `st0012-debug-test` branch
3. Open `test/document_test.rb` in `ruby-lsp`
4. Above the test class and methods, you should see 2 code lenses: `Test` and `Debug`
5. Set a breakpoint on line `14`
6. Click the `Debug` code lens above `test_valid_incremental_edits`
7. The test should be run in debug mode, and the debugger should stop at the breakpoint. Example:

    Screenshot 2023-04-26 at 18 25 42

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -264,7 +264,11 @@ export default class Client implements ClientInterface {
         Command.Update,
         this.updateServer.bind(this)
       ),
-      vscode.commands.registerCommand(Command.RunTest, this.runTest.bind(this))
+      vscode.commands.registerCommand(Command.RunTest, this.runTest.bind(this)),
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Works great!

PR #586, opened 2023-04-26
Add Formatter status item (v2)

andyw8:
To mitigate some of the potential complications in https://github.com/Shopify/vscode-ruby-lsp/pull/549/files, this is a simpler Status Item which shows only the configured `formatter` setting and does not attempt to determine the formatter which would be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,13 @@ vscode.commands.registerCommand(
 );
 ```
 
+## Formatting
+
+When `rubyLsp.formatter` is set to `auto`, Ruby LSP tries to determine which formatter to use.
+
+If the bundle has a **direct** dependency on a supported formatter, such as `rubocop` or `syntax_tree`, that will be used.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -189,6 +189,13 @@ vscode.commands.registerCommand(
 );
 ```
 
+## Formatting
+
+When `rubyLsp.formatter` is set to `auto`, Ruby LSP tries to determine which formatter to use.
+
+If the bundle has a **direct** dependency on a supported formatter, such as `rubocop` or `syntax_tree`, that will be used.
```
andyw8, line 196: (this will apply only after https://github.com/Shopify/ruby-lsp/pull/654 is merged).

egiurleo (approved):

egiurleo (approved):

egiurleo:
I liked it so much that I approved it twice 😆

PR #596, opened 2023-05-02
Add `Run In Terminal` code lens

st0012:
### Motivation

Because the test controller isn't able to handle dynamically generated tests, the `Run` code lens is not able to run them. This means those tests are not supported by any code lens anymore, while in prior versions they could be run in the terminal.

This PR adds a new code lens `Run In Terminal` that will run the tests in the terminal.

(This requires https://github.com/Shopify/ruby-lsp/pull/686)

### Implementation

1. Add terminal handling logic back to `client.ts`
2. Add a `runTestInTerminal` function that runs the code lens command in the terminal

Result: 

Screenshot 2023-05-02 at 23 18 11

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -286,22 +286,8 @@ export default class Client implements ClientInterface {
       vscode.commands.registerCommand(
         Command.Update,
         this.updateServer.bind(this)
-      ),
-      vscode.commands.registerCommand(
-        Command.DebugTest,
-        this.debugTest.bind(this)
       )
     );
-  }
-
-  private debugTest(_path: string, _name: string, command: string) {
-    return vscode.debug.startDebugging(undefined, {
-      type: "ruby_lsp",
-      name: "Debug",
-      request: "launch",
-      program: command,
-      env: this.ruby.env,
-    });
   }
 
   private async setupCustomGemfile() {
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -304,6 +313,14 @@ export default class Client implements ClientInterface {
     });
   }
 
+  private runTestInTerminal(_path: string, _name: string, command: string) {
```
vinistock, line 316: Let's encapsulate all of this test related stuff in the `TestController` class

bitwise-aiden (commented):
nit: Should we switch debug and run in terminal order?

st0012:
@bitwise-aiden Thanks for the suggestion, I will open a PR on `ruby-lsp` for it 👍

src/client.ts
```diff
@@ -289,21 +289,15 @@ export default class Client implements ClientInterface {
       ),
       vscode.commands.registerCommand(
         Command.DebugTest,
-        this.debugTest.bind(this)
+        this.testController.debugTest.bind(this)
+      ),
+      vscode.commands.registerCommand(
+        Command.RunTestInTerminal,
+        this.testController.runTestInTerminal.bind(this)
```
vinistock, line 296: Let's register the commands in the test controller as well. That way everything is encapsulated in the same class.

Also, this is currently binding a function from the `TestController` class to a `Client` instance, which could potentially lead to weird issues.

st0012: Oops I didn't know what I was actually doing 😅 Updated it now 👍

vinistock (approved):
🚀

PR #604, opened 2023-05-09
Remove unnecessary check for `test_` in test name

andyw8:
I recommend viewing with whitespace changes disabled:
https://github.com/Shopify/vscode-ruby-lsp/pull/604/files?w=1

### Motivation

This is to support https://github.com/Shopify/ruby-lsp-rails/pull/67

For minitest or test/unit tests, we expect the name in the form of `test_foo_does_something`. But for Rails tests, we expect the form `foo does something`.

Currently, the Test Explorer requires that the test name begins `test_` and so does not work with Rails tests.

But this check seems unnecessary. We are already checking in the gem that the test name begins with `test_`: https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_lsp/requests/code_lens.rb

### Implementation

Remove the check.

### Manual Tests

Open a test file and view the test explorer. Add a public method which does not begin with `test_`. Ensure it does not appear in the test explorer.

### DIFF
--- a/src/testController.ts
+++ b/src/testController.ts
@@ -16,6 +16,7 @@ export class TestController {
   private workingFolder: string;
   private terminal: vscode.Terminal | undefined;
   private ruby: Ruby;
+  private minitestTag = new vscode.TestTag("minitest");
 
   constructor(
     context: vscode.ExtensionContext,
@@ -40,7 +41,7 @@ export class TestController {
       },
       true
     );
-    this.testRunProfile.tag = new vscode.TestTag("minitest");
+    this.testRunProfile.tag = this.minitestTag;
 
     vscode.commands.executeCommand("testing.clearTestResults");
     vscode.window.onDidCloseTerminal((terminal: vscode.Terminal): void => {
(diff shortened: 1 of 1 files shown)


src/testController.ts
```diff
@@ -142,28 +142,26 @@ export class TestController {
         continue;
       }
 
-      if (test.id.startsWith("test_")) {
```
paracycle, line 145: Wasn't this check here to disambiguate minitest/test_unit tests from other kinds of tests, like RSpec tests? If so, we should make sure we mark tests properly as minitest/test_unit/rspec etc to be able to check appropriately here before removing this. Otherwise the output parsing code below might fail in a really bad way.

andyw8: I'm made an update which pass that additional information, and then uses tags to determine the behaviour. The problem is that this would break for older versions of the gem, so I'm wondering if we should [check for a minimum version](https://github.com/Shopify/ruby-dev-exp-issues/issues/921) of the gem.

cc @shopify-aiden, @bitwise-aiden

bitwise-aiden: Checking for a minimum version makes a lot of sense to me. We want to make sure to keep the two in sync in the long run, so adding this feature now will help make steps toward that 👍🏻

PR #605, opened 2023-05-09
Add gem actions in the Gemfile

snutij:
### Motivation

See: https://github.com/Shopify/ruby-lsp/issues/585
Link to: https://github.com/Shopify/ruby-lsp/pull/710

### Implementation

- Register the new command `rubyLsp.openGem`
- When the command is triggered open the link through `openExternal`

I had to filter lenses before calling `createTestItems` because it can crash now that we receive two different kinds of codelens which are not formatted in the same way. For this reason, this pull request should be merged **before** server one.

### Automated Tests

I don't know how to test it. But feel free to point me in the right direction 🙏 

### Manual Tests

- Check out this branch
- Checkout ruby-lsp branch 
- Open a Gemfile
- Something like that is expected:

<details><summary>Record</summary>

Screen-Recording-2023-05-09-at-22 37 45
</details>


src/client.ts
```diff
@@ -310,6 +312,10 @@ export default class Client implements ClientInterface {
       vscode.commands.registerCommand(
         Command.Update,
         this.updateServer.bind(this)
+      ),
+      vscode.commands.registerCommand(
+        Command.OpenGemHomepage,
+        this.openGemHomepage.bind(this)
```
vinistock, line 318: NIT: let's rename both the command and function to be `openLink` so that it's more generic. We may want to link to other places in the editor in the future and then we can reuse this function.

vinistock (approved):

PR #629, opened 2023-05-31
Drop Ruby 2.7 support

andyw8:
### Motivation

Following up from https://github.com/Shopify/ruby-lsp/pull/732


andyw8:
After investigating with Vini, this is not actually needed. For users who remain 2.7, they will get the latest version of the gem which supports 2.7, i.e. 0.5.1.

PR #631, opened 2023-06-05
Support env on debug launch configurations

### LINKED ISSUE #630: Support env in launch configuration
### This feature is VS Code specific

- [X] VS Code specific

### Use case

Some projects use env var to change program or test behaviour, like how many processes to be used. Therefore we need to let users specify those env vars in `launch.json`.

### Description

Allow users to specify `env` when defining actions in `launch.json`, like:

```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "ruby_lsp",
            "request": "launch",
            "name": "Debug test file",
            "program": "ruby -Itest ${relativeFile}",
            "env": { "FOO": "BAR" }
        },
    ]
}
```

### Implementation

_No response_

rafaelfranca:
### Motivation

In the Rails project we need to pass an environment variable to disable parallel running of the tests. This commit adds support to set this configuration in the debug launch configuration instead of requiring the user to change the test runner script.

Fixes #630.

### DIFF
--- a/package.json
+++ b/package.json
@@ -204,6 +204,13 @@
               "program": {
                 "type": "string",
                 "description": "The program to debug"
+              },
+              "env": {
+                "type": "object",
+                "description": "Environment variables defined as a key value pair. Example: { \"BACKTRACE\": \"1\" }",
+                "additionalProperties": {
+                  "type": "string"
+                }
               }
             }
           },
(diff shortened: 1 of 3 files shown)


st0012 (approved):
Tested and was working well 👍 
I will open an issue to see how to make code lens pick up launch configurations as a follow up.

PR #632, opened 2023-06-06
Add devcontainer configuration

rafaelfranca:
To test this repository a lot of conditions need to be met.

* You need ruby installed
* You need ruby installed with YJIT
* In order to satisfy the above you need rust installed

This commit adds a devcontainer configuration that will install all those dependencies and leave the project configured so you can just run the tests and get a green build.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,28 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+  "name": "Node.js & TypeScript",
+  // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:0-20",
+  "features": {
+    "ghcr.io/devcontainers/features/github-cli:1": {},
+    "ghcr.io/devcontainers/features/rust:1": {},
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+{
+  "name": "Node.js & TypeScript",
+  // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:0-20",
+  "features": {
+    "ghcr.io/devcontainers/features/github-cli:1": {},
+    "ghcr.io/devcontainers/features/rust:1": {},
+    "ghcr.io/devcontainers/features/ruby:1": {}
+  },
+  "overrideFeatureInstallOrder": ["ghcr.io/devcontainers/features/rust"]
```
bitwise-aiden, line 12: For my info: The docs aren't specific on this feature, by just supplying rust, does that bump it above the others and they will run afterwards?

rafaelfranca: Yes. https://github.com/devcontainers/spec/blob/c0beafa8f08db7df99ca12af9d407fb6c9b5f8db/docs/specs/devcontainer-features.md?plain=1#L270

bitwise-aiden: Weird that this isn't propagated to their [JSON reference](https://containers.dev/implementors/json_reference/). Thanks for that!

PR #639, opened 2023-06-10
Use testItem kind

andyw8:
### Motivation

https://github.com/Shopify/ruby-lsp/pull/734 added a `kind` to test items, of either `example` or `group`. The extension can use this to distinguish running a single tests from a full test file.

Previously we were doing this by checking for `test_` at the beginning of the name. But that approach isn't viable for some test frameworks, e.g. Rails declarative tests, or RSpec, which other Ruby LSP extensions may need to support.

### Implementation

Since we don't own the TestItem interface, there are limited way to supplement it with this additional information. I went with use a `TestTag`.

### Automated Tests

Nothing yet, it could be tricky. I'd like to collect some opinions on how to approach this.

### Manual Tests

Note that this relies on behaviour which in `main` of Ruby LSP but not yet released.

To ensure the new tag is being read, and it isn't just falling back to the previous behaviour, I suggest temporarily changing the two `test_` checks to `notest_` in `testController.ts`.

* Start debugging this branch, and open `ruby-lsp`.
* Open a test file contains multiple tests, e.g. `test/rubocop/cop/ruby_lsp/use_language_server_aliases_test.rb`.
* Click on the class 'Run in Terminal' and verify it runs the whole test on the command line.
* Click on the class 'Run' and verify it runs the whole test on the command line.
* Click on the individual test 'Run' and verify is run in the Test Explorer.
* Set a breakpoint by clicking the gutter. Click 'Debug' and ensure it pauses.


src/testController.ts
```diff
@@ -106,8 +110,14 @@ export class TestController {
 
       // Add test methods as children to the test class so it appears nested in Test explorer
       // and running the test class will run all of the test methods
-      if (name.startsWith("test_")) {
-        testItem.tags = [this.debugTag];
+
+      // Older Ruby LSP versions may not include 'kind' so we try infer it from the name.
+      if (!res.data?.kind && name.startsWith("test_")) {
+        res.data.kind = "example";
+      }
```
st0012, line 117: Can we move this part to the `else` branch of the above `if (res.data?.kind) {` condition, and insert an `example` tag there?

src/testController.ts
```diff
@@ -106,8 +110,14 @@ export class TestController {
...
+
+      if (res.data?.kind === "example") {
```
st0012, line 119: If we always assign `kind` to tags in the above conditions, we can simplify this to just check the first tag instead of `res.data`?

andyw8: I've updated it to check the tags instead of `res.data`, but relying on its position in the array seems fragile so I've used a `find`.

src/testController.ts
```diff
@@ -106,8 +110,14 @@ export class TestController {
 
       // Add test methods as children to the test class so it appears nested in Test explorer
       // and running the test class will run all of the test methods
-      if (name.startsWith("test_")) {
-        testItem.tags = [this.debugTag];
+
+      // Older Ruby LSP versions may not include 'kind' so we try infer it from the name.
+      if (!res.data?.kind && name.startsWith("test_")) {
```
st0012, line 115: Why can't this be a `else if` branch of the `if (res.data?.kind) {` condition?

src/testController.ts
```diff
@@ -106,8 +110,14 @@ export class TestController {
...
+        res.data.kind = "example";
```
st0012, line 116: Do you mean adding it to tags? I thought the idea was to add `"example"` to tags and only check tags later?

andyw8:
(I'd still like to add tests around this later).

PR #649, opened 2023-06-16
Send configuration telemetry

vinistock:
### Motivation

Send configuration telemetry to the private metrics extension on activation. The idea is to capture configuration related to the Ruby LSP once a day, so that we can understand how people are using it.

### Implementation

1. Added a new method `sendConfigurationEvents` to `Telemetry`, which reads all configurations we list and sends it to the private extension. The method only does it once a day
2. Started invoking the method on activation

### Automated Tests

Added a test to verify that the events are being sent.

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -1,6 +1,6 @@
 import * as vscode from "vscode";
 
-export interface TelemetryEvent {
+interface RequestEvent {
   request: string;
   requestTime: number;
   lspVersion: string;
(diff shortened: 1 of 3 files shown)


src/telemetry.ts
```diff
@@ -37,8 +50,57 @@ export class Telemetry {
...
+        field,
+        value: (
+          vscode.workspace.getConfiguration(namespace).get(field) ?? ""
+        ).toString(),
+      });
+    });
+
+    const enabledFeatures = vscode.workspace
+      .getConfiguration("rubyLsp")
+      .get("enabledFeatures")!;
```
andyw8, line 86: Isn't there a chance `get("enabledFeatures")` could be null?

vinistock: I don't think so because we have a default value configured. It always returns an object.

andyw8 (approved):

Issue #381, opened 2023-02-08
Support using the LSP with multiple projects in the same VS Code instance


vinistock:
VS Code allows opening multiple projects simultaneously in the same instance. For these scenarios, we can't spawn the Ruby LSP at the top level.

Each project may be using different versions of the LSP or RuboCop and may also have different RuboCop configurations. For these scenarios, it may be necessary to spawn multiple servers, one for each project.

vinistock:
I have made a prerelease with multi-workspace support https://github.com/Shopify/vscode-ruby-lsp/releases/tag/v0.5.0.

If people want to give it a try, please provide feedback on this issue so that we can address it before a stable release.

You can install prerelease version of VS Code extensions by

- Going into the extensions tab
- Finding the extension you want to install a prerelease version
- Clicking the squares button to the left of the cog icon

joelmoss:
Just tried and its working perfectly! thx loads!

vinistock:
Thank you for the feedback! The latest commit on the PR addresses that issue.

jprosevear:
@vinistock I'm a bit unclear how this solves the issue that was solved in #877 - if i open the root directory of a project:

```
monorepo/
  client/package.json...
  server/Gemfile
```
 
`this.workingDirectory` directory is still `monorepo/` the formatter is not automatically detected and the .ruby-lsp dir is written at the root.

Is this patch intended only for multi root workspace scenarios?

(tested with pre-release version and debugged with 0.5.0 from code locally)

vinistock:
You need to open `client` and `server` as separate workspaces. Don't open the top level folder, open `client` and then add `server` as a second workspace.

jprosevear:
For future readerrs, here is the setup working for me (with 0.5.0) - `mono.code-workspace`:

``` 
{
	"folders": [
		{
			"name": "root",
			"path": "./"
		},
		{
			"path": "client"
		},
		{
			"path": "server"
		},
	],
	"settings": {
		"files.exclude": {
			"client/**": true,
			"server/**": true
		}
	}
}
```

This prevents duplication of folders in "root" (there are a few files there).

This still installs .ruby-lsp in both root and server.

However if I change the order of the folders in the workspace so that root is last, format on save breaks.

vinistock:
Please see my answer in https://github.com/Shopify/vscode-ruby-lsp/pull/877#issuecomment-1839262550.

PR #655, opened 2023-06-20
Add `customWorkspaceFolder` setting to set a custom `workingFolder`

nicolasrouanne:
### Motivation

Closes #621 

Allow opening a VS Code workspace in any folder (not only those containing a `Gemfile.lock` at the root), and have `ruby-lsp` properly run.
 > Only mandatory condition is that there is a `Gemfile.lock` in a subfolder of the VS Code workspace.

### Implementation

I added a `workingFolder` in `Client` constructor (was already present in `Ruby`, `Debugger` and `TestController`). If `customWorkspaceFolder` configuration is set, initialize all classes with it, otherwise fallback to default VSCode workspace.

> it's probably possible to extract this `workingFolder` resolution out of `extension.ts`, but I wanted to see if the approach was accepted first.

### Automated Tests

This is merely at proof of concept state. I'll be happy to add some tests if the implementation is deemed relevant.

### Manual Tests

- I added the `customWorkspaceFolder` setting in my VS Code instance.
- I ran `yarn package` to build a VSIX file
- I installed the extension through "Install from VSIX..."

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   ],
   "activationEvents": [
     "onLanguage:ruby",
-    "workspaceContains:Gemfile.lock"
+    "workspaceContains:**/Gemfile.lock"
   ],
   "main": "./out/extension.js",
   "contributes": {
@@ -178,6 +178,11 @@
         },
         "rubyLsp.bundleGemfile": {
           "description": "Relative path to the Gemfile to use for bundling the Ruby LSP server. Only necessary when using a separate Gemfile for the Ruby LSP",
+          "type": "string",
+          "default": ""
+        },
+        "rubyLsp.customWorkspaceFolder": {
+          "description": "A custom working directory for the Ruby LSP server. Only necessary if your Gemfile isn't located at the VS Code workspace root",
           "type": "string",
           "default": ""
         }
--- a/src/client.ts
+++ b/src/client.ts
@@ -46,9 +46,10 @@ export default class Client implements ClientInterface {
     context: vscode.ExtensionContext,
     telemetry: Telemetry,
     ruby: Ruby,
-    testController: TestController
+    testController: TestController,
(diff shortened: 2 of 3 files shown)


nicolasrouanne:
I have signed the CLA!

nicolasrouanne:
I understand the added complexity of having yet another setting, only for this specific usecase, that can be addressed by a more generic way (multi workspace). I agree it's probably not worth it in the long run, considering the added pain of maintaining it.

That made me think, maybe there's another way which would solve #621 without any additional setting: what if we set the `currentWorkingDirectory` to the first (most upper one in directory tree) location where a `Gemfile.lock` is found instead of the workspace root?
- it would be a non breaking change, as it would default to `workspace root`
- it wouldn't require any setting to maintain
- it would work for simple _frontend_ / _backend_ monorepo usecases

For most complex setups, the multi workspace is still the best solutions.

vinistock:
In terms of the complexity, it's not just the new setting, but all the code related to it. It's extra complexity in the codebase to fix an inconvenience: not having to open two instances of VS Code (one for the frontend and another one for the backend) because the codebase is split.

After a change like that makes its way into the codebase, everything we add will need to be careful about alternative project structures and take that into account.

I also feel like searching for `Gemfile.lock` in subfolders can be brittle and lead to more complexity in the future. What if we find a `Gemfile.lock` that includes RuboCop, but `.rubocop.yml` doesn't exist in the same level? Do we search in levels above to find some config? Do we warn the user? What if there's a `.ruby-version` file in a subfolder but not on the top level? This and many other corner case scenarios would force us into taking a lot of complexity for a use case that's not common enough.

If we're going to take extra complexity on, then it's better to do it for #381, which is a more common use case and also solves #621.

nicolasrouanne:
Thanks for taking the time to thoroughly explain and justify your point of view @vinistock. Much appreciated 🙏🏻 
Closing this in favour of #381

PR #664, opened 2023-06-27
Remove custom bundle logic

### LINKED ISSUE #421: Move `.ruby-lsp` folder logic into the server
Based on https://github.com/Shopify/vscode-ruby-lsp/pull/419#discussion_r1120876546.

To make the Ruby LSP work similarly on every editor, we should move the `.ruby-lsp` logic into the server. This might require us to create another executable that basically just sets up the folder, sets up bundle and then just runs `bundle exec ruby-lsp` at the end.

### LINKED ISSUE #536: Using a custom path while using system gems is unsupported
The extension fails during initialization with the following output:

```
>> Running `bundle install`...
ERROR> Using a custom path while using system gems is unsupported.
ERROR> 
ERROR> path:
ERROR> Set via BUNDLE_PATH: "/bundle/vendor"
ERROR> 
ERROR> path.system:
ERROR> Set via BUNDLE_PATH__SYSTEM: true
ERROR> 
ERROR> disable_shared_gems:
ERROR> You have not configured a value for `disable_shared_gems`
>> `bundle install` exited with code 15
```

Environment:
ruby 3.2.1 (2023-02-08 revision 31819e82c8) [aarch64-linux]
vscode: 1.77.0
vscode-ruby-lsp: 0.2.3
bundler: 2.4.10

The environment runs inside a docker container using the environment variable `BUNDLE_PATH=/bundle/vendor`.
I use this environment variable to persist gems inside the vscode devcontainer.

Looking at the code, I can see that the BUNDLE_PATH__SYSTEM is always set to true. 
https://github.com/Shopify/vscode-ruby-lsp/blob/09691bd596be17574512adf5d1aebae79e6fe74f/src/ruby.ts#L210-L212

Am I missing something in the configuration? Does vscode-ruby-lsp support a custom BUNDLE_PATH in any way?

### LINKED ISSUE #550: Investigate using bundler compose to run server
The new [bundler-compose](https://github.com/segiddins/bundler-compose) plugin allows a gem executable to add the bundle to its LOAD_PATH, effectively solving the problem we use the custom bundle for.

If we can use it to avoid maintaining the custom bundle, we absolutely should. In my head, this is the easiest way to move forward with this:
1. Add functionality to bundler-compose to allow us to do this in the `ruby-lsp` executable
```ruby
# Equivalent of bundler/setup, but includes the Ruby LSP's dependencies and the bundle's dependencies in the LOAD_PATH
# If no bundle exists, only includes the Ruby LSP's dependencies in the LOAD_PATH
require "bundler/compose"
```
2. Add bundler-compose as a dependency of the Ruby LSP and add the require in `exe/ruby-lsp`
3. Remove all of the custom bundle logic and simply always run the Ruby LSP from the global installation (without bundle exec)
4. Use part of our logic just to automatically install and update the server gem (using gem install ruby-lsp or gem update ruby-lsp)

### LINKED ISSUE #638: Ensure custom lockfile stays in sync with `Gemfile.lock`
We've had reports about situations where Ruby LSP is still using an older RuboCop version after a RuboCop update.

As per [this thread](https://ruby-dx.slack.com/archives/C050GHEK6BU/p1686103572452229) in the public Ruby DX slack, we can store the hash of the original lockfile in the `.ruby-lsp/Gemfile` as a comment or, better, in another file under the `.ruby-lsp` folder so that we can more precisely determine if the lockfile has changed, instead of comparing file modification times.

This is similar to the [approach taken by bundler-compose](https://github.com/segiddins/bundler-compose/blob/9bed00a6d7e6baad7be4da6abf4f902f30cfc517/lib/bundler/compose/cli.rb#L58-L64).

### LINKED ISSUE #682: Bundle install fails if system installation directory is read-only
### Operating System

Ubuntu 20.04

### Ruby version

3.2.2

### Project has a bundle

- [X] Has bundle

rbenv

### Description

Ruby-lsp forces Bundler to install to the system RubyGems directory as shown in ruby.ts:

```typescript
// We must use the default system path for bundler in case someone has BUNDLE_PATH configured. Otherwise, we end
// up with all gems installed inside of the `.ruby-lsp` folder, which may lead to all sorts of errors
this._env.BUNDLE_PATH__SYSTEM = "true";
```

However this fails if the system RubyGems directory is read-only, e.g. when Ruby is installed via package manager. For example when using Fullstaq Ruby, the system location is something like /usr/lib/fullstaq-ruby/versions/3.2-jemalloc/lib/ruby/gems/3.2.0.

Normally setting `bundle config path ~/.gem` fixes the problem, but Ruby-lsp overrides this by setting BUNDLE_PATH__SYSTEM=true. Result: all sorts of permission errors.

```
>> Running `bundle install`...
ERROR> Retrying download gem from https://rubygems.org/ due to error (2/4): Bundler::PermissionError There was an error while trying to write to `/usr/lib/fullstaq-ruby/versions/3.2-jemalloc/lib/ruby/gems/3.2.0/cache/prettier_print-1.2.1.gem`. It is likely that you need to grant write permissions for that path.
...etc...
```

I don't know what "end up with all gems installed inside of the `.ruby-lsp` folder" means but I think you should either add a config option to disable BUNDLE_PATH__SYSTEM, or an option to force Bundler to install ~/.gem.

vinistock:
### Motivation

Closes #421
Closes #550
Closes #638
Closes #536
Closes #682

Complete explanation in https://github.com/Shopify/ruby-lsp/pull/789.

### Implementation

Removed everything related to setting up the custom bundle and started using `gem install` and `gem update` to manage the Ruby LSP server.


src/client.ts
```diff
@@ -461,164 +363,39 @@ export default class Client implements ClientInterface {
...
-      install.on("error", (error) => {
-        stdout.close();
-        stderr.close();
-        reject(error);
+    // If we haven't updated the gem in the last 24 hours, update it
+    if (
+      lastUpdatedAt === undefined ||
+      Date.now() - lastUpdatedAt > oneDayInMs
+    ) {
+      await asyncExec("gem update ruby-lsp", {
```
st0012, line 393: Can we have some fault-tolerant mechanism here so it doesn't break when the user has no internet connection?

st0012: Unfortunately the code I got was `1` 😅 

```
$ gem update ruby-lsp
ERROR:  While executing gem ... (Gem::RemoteFetcher::FetchError)
    SocketError: Failed to open TCP connection to rubygems.org:443 (https://rubygems.org/specs.4.8.gz) (Gem::RemoteFetcher::FetchError)
        /opt/rubies/3.2.2/lib/ruby/3.2.0/rubygems/remote_fetcher.rb:261:in `rescue in fetch_path'
        /opt/rubies/3.2.2/lib/ruby/3.2.0/rubygems/remote_fetcher.rb:241:in `fetch_path'
...

$ echo $?            
1
```

vinistock: Well, thinking more about it, we don't want to prevent the server from starting on any `gem update` errors. I added a `try catch` block to log when something goes wrong.

st0012 (approved):

vinistock:
Some questions to help understand:
1. Is RuboCop is your Gemfile?
2. Was a `.ruby-lsp/Gemfile` generated for you?
3. Any messages in the output tab?

vinistock:
Can you please try this and report back what happens?
1. Remove the `ruby-lsp` from the `Gemfile` again
2. Make sure to bundle install to reflect the changes in `Gemfile.lock`
3. Run `ruby-lsp` (no `bundle exec`) manually in the terminal

Do you get anything printed?

krakowski:
I switched to the prerelease version of `vscode-ruby-lsp` in order to get this fix since I also encountered #536. Using this fix, Ruby LSP seems to hang during startup.

Screenshot from 2023-07-25 14-44-17

Running it manually in the terminal (without bundler) just prints `Starting Ruby LSP...` (after installing all gems from my `Gemfile`). After Killing the server the generated stacktrace shows, that it was stuck at `ruby_lsp/server.rb:65:in 'start'`.

xeron:
Just tested with 0.4.2 beta with `ruby-lsp` removed from the app's Gemfile and rubocop linting is working. Thanks!

xeron:
Well it still creates `.ruby-lsp` with a Gemfile which does eval + adds ruby-lsp gem. Not sure if this is intended but now every time I do `bundle update && bundle clean` it removes lsp from the local bundle setup:

```
$ bundle update && bundle clean
Fetching gem metadata from https://rubygems.org/..........
Resolving dependencies...
Bundle updated!
Removing prettier_print (1.2.1)
Removing ruby-lsp (0.7.4)
Removing sorbet-runtime (0.5.10932)
Removing syntax_tree (6.1.1)
```

xeron:
I see. Thanks. I guess I got the wrong impression that `ruby-lsp` gem won't be needed in neither `Gemfile` nor `.ruby-lsp/Gemfile` after this PR.

PR #667, opened 2023-06-29
Execute all test controller items in one go

vinistock:
### Motivation

Currently, each test item in the controller is executed in a separate process, which is a bit wasteful given that we only display the tests for the current file.

We can instead execute all tests in a single command and then just parse the output to show to the user.

### Implementation

The basic idea is:
1. Flatten the test list. Initially, the test items are nested (i.e.: test methods are inside classes). We need a flat list that only includes the specific examples that will be executed
2. Then we generated the command to execute all selected examples in one go. For Minitest and test-unit, that means creating a big regex that will match all selected examples. For Rails, it just means adding all paths and line numbers (e.g.: `test/foo_test.rb:5 test/foo_test.rb:15`...)
3. If all tests pass, then we just update that they all passed. If something fails, then we need to parse the output, match the failure message with the test item and only fail those

### Manual Tests

Suggested repos:
- `ruby-lsp` -> minitest
- `ruby-lsp-rails` -> rails
- `yarp` -> test-unit

1. Open any test file
2. Execute tests, verify that all results show up simultaneously
3. Introduce an intentional failure
4. Run the tests
5. Verify the failure shows up with the correct message

### DIFF
--- a/src/testController.ts
+++ b/src/testController.ts
@@ -11,7 +11,11 @@ const asyncExec = promisify(exec);
 
 export class TestController {
   private testController: vscode.TestController;
-  private testCommands: WeakMap<vscode.TestItem, string>;
+  private testCommands: WeakMap<
+    vscode.TestItem,
+    { command: string; identifier: string }
+  >;
+
   private testRunProfile: vscode.TestRunProfile;
   private testDebugProfile: vscode.TestRunProfile;
   private debugTag: vscode.TestTag = new vscode.TestTag("debug");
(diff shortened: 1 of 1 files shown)


src/testController.ts
```diff
@@ -91,37 +98,44 @@ export class TestController {
 
     response.forEach((res) => {
       const [_, name, command, location] = res.command!.arguments!;
-      const testItem: vscode.TestItem = this.testController.createTestItem(
-        name,
-        name,
-        uri
-      );
+      let testItem;
```
KaanOzkan, line 101: Getting `Variable 'testItem' implicitly has type 'any' in some locations where its type cannot be determined.` locally from TS lsp.

KaanOzkan (commented):
I can't get the failures to show up when I run all tests in a file. For both exceptions and incorrect assertions.

Screenshot 2023-06-30 at 3 14 35 PM

vinistock:
This is quite complicated to get right and feels a bit too rigid. I'm closing this for now.

PR #669, opened 2023-06-30
Enable code lens by default

vinistock:
### Motivation

To pair with https://github.com/Shopify/ruby-lsp/pull/792, let's enable code lens by default now that they're a stable feature.


andyw8:
@vinistock when are you hoping to ship this? It might be good to get https://github.com/Shopify/shopify/pull/432693 in first, and allow a few days for some selected people to try it.

bitwise-aiden (approved):

PR #677, opened 2023-07-04
Fix fetching server version for configured bundle path

### LINKED ISSUE #671: Support custom bundle gemfile when checking server version for telemetry
### Operating System

macOS 13.4.1 (22F82)

### Ruby version

2.7.4

### Project has a bundle

- [X] Has bundle

rvm

### Description

Extension fails to activate if ruby-lsp is not included in root `Gemfile` or no `.ruby-lsp/Gemfile` is present, which is a legit scenario when `"rubyLsp.bundleGemfile"` is set to a custom path.

I worked this around by adding a hacky `.ruby-lsp/Gemfile`:
```ruby
eval_gemfile(File.expand_path("../local_gems/Gemfile", __dir__))
gem "debug", require: false, group: :development, platforms: :mri, source: "https://rubygems.org"
```
Where `local_gems/Gemfile` is the same value I set for `"rubyLsp.bundleGemfile"`.

vinistock:
### Motivation

Closes #671

If the user has configured a custom Gemfile outside of the project, then we need to use that to check the version of the server.

### Implementation

Added another check to define the correct `bundleGemfile` to use.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -624,8 +624,17 @@ export default class Client implements ClientInterface {
 
   private async getServerVersion(): Promise<string> {
     let bundleGemfile;
+    const customBundleGemfile: string = vscode.workspace
+      .getConfiguration("rubyLsp")
+      .get("bundleGemfile")!;
 
-    if (fs.existsSync(path.join(this.workingFolder, ".ruby-lsp", "Gemfile"))) {
+    // If a custom Gemfile was configured outside of the project, use that. Otherwise, prefer our custom bundle over the
+    // app's bundle
+    if (customBundleGemfile.length > 0) {
+      bundleGemfile = `BUNDLE_GEMFILE=${customBundleGemfile}`;
+    } else if (
+      fs.existsSync(path.join(this.workingFolder, ".ruby-lsp", "Gemfile"))
+    ) {
       bundleGemfile = `BUNDLE_GEMFILE=${path.join(
         this.workingFolder,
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -624,8 +624,17 @@ export default class Client implements ClientInterface {
 
   private async getServerVersion(): Promise<string> {
     let bundleGemfile;
-
-    if (fs.existsSync(path.join(this.workingFolder, ".ruby-lsp", "Gemfile"))) {
+    const customBundleGemfile: string = vscode.workspace
+      .getConfiguration("rubyLsp")
+      .get("bundleGemfile")!;
+
+    // If a custom Gemfile was configured outside of the project, use that. Otherwise, prefer our custom bundle over the
```
bitwise-aiden, line 631: > configured outside of the project

Is this at odds with using `. getConfiguration("rubyLsp")`?

vinistock: No, by outside the project we mean outside of the current working directory. Some users like to have a separate `Gemfile` only for development dependencies and we provide this configuration as a way to point to it.

For example, something like this
```shell
# Regular bundle
project/Gemfile

# Ruby LSP custom bundle
project/.ruby-lsp/Gemfile

# User configured custom bundle
~/some/other/path/outside/of/project/Gemfile
```

KaanOzkan (approved):

bitwise-aiden (approved):

PR #680, opened 2023-07-05
Show errors when working on ruby-lsp and ruby-lsp-rails

### LINKED ISSUE #585: Surface errors when developing `ruby-lsp` or `ruby-lsp-rails`
Depends on #571 

When errors occur in the server, we send them to telemetry if a client is defined by a private extension using the `ruby-lsp.getPrivateTelemetryApi` command.

However, when we're working on the `ruby-lsp` or `ruby-lsp-rails` themselves, we should shorten this feedback loop and just display the errors directly in the window.

Let's start checking if we're working on these two projects and then, instead of sending telemetry on errors, we just use `vscode.window.showErrorMessage` to pop up the error to immediately bring it to our attention.

vinistock:
### Motivation

Closes #585

When working on `ruby-lsp` and `ruby-lsp-rails`, it's better to have errors surfaced straight away than sending to telemetry, so that we can see it and address it soon. This will be particularly useful for large migrations, like transitioning to YARP.

### Implementation

Started showing a dialogue if the error telemetry occurred inside of `ruby-lsp` or `ruby-lsp-rails`.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -202,13 +202,25 @@ export default class Client implements ClientInterface {
       clientOptions
     );
 
-    this.client.onTelemetry((event) =>
-      this.telemetry.sendEvent({
-        ...event,
-        rubyVersion: this.ruby.rubyVersion,
-        yjitEnabled: this.ruby.yjitEnabled,
-      })
-    );
+    const baseFolder = path.basename(this.workingFolder);
+
+    this.client.onTelemetry((event) => {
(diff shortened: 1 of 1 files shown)


egiurleo (commented):
This is a good idea and I really appreciate how this will tighten our feedback loops 💪🏻

Is there any value in sending these errors to telemetry in addition to showing the error in VSCode? I'm assuming most devs are not working on ruby-lsp or ruby-lsp-rails, so the errors from these repos are likely a drop in the bucket for us, but I imagine there's still value from capturing them in our telemetry?

PR #684, opened 2023-07-10
Add show syntax tree request

vinistock:
### Motivation

Server part https://github.com/Shopify/ruby-lsp/pull/805

Add a command to show the syntax tree of the current Ruby document.

### Implementation

The new command sends a custom request to the server to get the pretty printed AST back.

Then we use a custom `DocumentProvider` to handle a new `ruby-lsp://` URI scheme. The only reason this is needed is so that the opened file doesn't have to be saved by the user and can be closed directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,10 @@
       {
         "command": "rubyLsp.update",
         "title": "Ruby LSP: Update language server gem"
+      },
+      {
+        "command": "rubyLsp.showSyntaxTree",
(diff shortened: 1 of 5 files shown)


vinistock:
There's no element in the UI that shows it. You open the command palette with `CMD + SHIFT + P` and search for `Ruby LSP: Show syntax tree`.

bitwise-aiden (approved):

PR #691, opened 2023-07-12
Add selection support for show syntax tree

vinistock:
### Motivation

Server side PR https://github.com/Shopify/ruby-lsp/pull/810

Checking the AST for the entire file is useful, but sometimes the developer might want to know only the AST for a specific selection. This PR makes our command handle the current selection to only display the AST for that part.

### Implementation

Selections in VS Code use an `anchor` (the initial point where the user started dragging) and `active` (the final point of the selection). If nothing is selected, both still exist, but they have the exact same location.

So, if `anchor` and `active` have different positions, it means there's an actual selection and we can instantiate a range and send it as a request parameter.

**Note**: if the user selects from bottom up, then the positions are inverted (`active` will happen before `anchor`). In that case, `isReversed` is `true` and we need to account for it when instantiating the range.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -10,6 +10,7 @@ import {
   Executable,
   RevealOutputChannelOn,
   CodeLens,
+  Range,
 } from "vscode-languageclient/node";
 
 import { Telemetry } from "./telemetry";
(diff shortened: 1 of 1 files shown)


andyw8 (approved):
Verified against https://github.com/Shopify/ruby-lsp/pull/810

PR #696, opened 2023-07-13
Expose test runner commands

paracycle:
### Motivation

If the extension exposes test runner commands (`Run`, `Run in Terminal` and `Debug`), then users can execute those via the command palette or assign keybindings for them. This PR is doing exactly that.

### Implementation

When we expose the commands as runnable items, that means a command could be run with no arguments. In the current state, since we always run the commands via the code lenses, that isn't a problem. So, the tricky part of this PR is finding the right test to run in that case.

The fallback is to find the test item that matches the active line in the active text editor and run that.

### Automated Tests

No automated tests

### Manual Tests

While on this branch:
1. Launch a Ruby project
2. Open a test file, and place the cursor inside a test case.
3. In the Command Palette, run `Ruby LSP: Run current test`, or `Ruby LSP: Run current test in terminal`, or `Ruby LSP: Debug current test` to see the active test case run.

To make sure the command fails gracefully, you can test the following cases where none of them should show an error nor cause a test to run:
1. There is no active editor
2. The active editor is not a Ruby one
3. The active editor selection is not inside a test case


paracycle:
> I couldn't get the debug command to work though. The code lens still does, but the command always exits with status 1.

I'll check that and fix, thanks for checking...

paracycle:
@vinistock it turns out I forgot to add the code for that code path, [now](https://github.com/Shopify/vscode-ruby-lsp/pull/696/files#diff-543a0fe15896a7a599e39cc1f1da3b647e430c1819605e520624e60e4a13b062R141-R143) this should be fixed.

PR #697, opened 2023-07-13
Stop manually prefixing commands with category

paracycle:
### Motivation

Streamline the extension

### Implementation

Extract `Ruby LSP:` prefixes to `"Ruby LSP"` category for each command.

### Automated Tests

No automated tests.

### Manual Tests

Launch the extension on this branch and see that the command titles in the Command Palette are not changed

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,23 +35,28 @@
     "commands": [
       {
         "command": "rubyLsp.start",
-        "title": "Ruby LSP: Start"
+        "title": "Start",
+        "category": "Ruby LSP"
       },
       {
         "command": "rubyLsp.restart",
-        "title": "Ruby LSP: Restart"
+        "title": "Restart",
+        "category": "Ruby LSP"
       },
       {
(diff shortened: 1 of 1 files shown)


st0012:
@vinistock while testing for this PR, I found that the `Show syntax tree` command doesn't do anything on my machine. But the request was made and responded:

```
[Trace - 16:43:22] Sending request 'rubyLsp/textDocument/showSyntaxTree - (147)'.
[Trace - 16:43:22] Received response 'rubyLsp/textDocument/showSyntaxTree - (147)' in 1ms.
```

vinistock (approved):
Nice. Thank you for the fix!

vinistock:
@st0012 are you on the `main` branch of the LSP? Show syntax tree is not released yet.

It responds still because we return `nil` by default for requests we don't handle.

st0012:
@vinistock Ah I see 🤦‍♂️  Thanks for clarification!

PR #708, opened 2023-07-18
Add default keybinding for running test

andyw8:
### Implementation

The default is based on dev-test-runner (an internal Shopify extension).


andyw8:
We will use the VS Code defaults instead:

https://github.com/Shopify/vscode-ruby-lsp/pull/797

PR #716, opened 2023-07-23
allow to use the extension in folders that contain spaces

tonklon:
### Motivation

The extension does not work for projects in folders, were there is a space somewhere in the path. I often use that and I don't want to rename all my project folders.

### Implementation

Just adding spaces around the environment Variable named BUNDLE_GEMFILE is enough.

`BUNDLE_GEMFILE="/path with spaces/project/Gemfile" bundle exec ruby-lsp` works just fine, while
`BUNDLE_GEMFILE=/path with spaces/project/Gemfile bundle exec ruby-lsp` won't work

### Automated Tests

No, I have not included test for this. I was just about writing the tests and working on for https://github.com/Shopify/ruby-lsp/pull/781 when I recognized the extension itself was broken, too.

I might perhaps come back after the ruby-lsp itself works with spaces and add some. Meanwhile I guess the change is quite simple, and since I need to do it, to be able to work on the other issue, I could as well publish the code and add a PR, if somebody else has this.

### Manual Tests

Add a space to the folder your test project resides in, and try to start ruby-lsp. It will fail. After this change ruby-lsp can start.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -433,20 +433,20 @@ export default class Client implements ClientInterface {
     // If a custom Gemfile was configured outside of the project, use that. Otherwise, prefer our custom bundle over the
     // app's bundle
     if (customBundleGemfile.length > 0) {
-      bundleGemfile = `BUNDLE_GEMFILE=${customBundleGemfile}`;
+      bundleGemfile = `BUNDLE_GEMFILE="${customBundleGemfile}"`;
     } else if (
       fs.existsSync(path.join(this.workingFolder, ".ruby-lsp", "Gemfile"))
     ) {
-      bundleGemfile = `BUNDLE_GEMFILE=${path.join(
+      bundleGemfile = `BUNDLE_GEMFILE="${path.join(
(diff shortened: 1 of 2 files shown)


Morriar:
Should we move this as draft since it won't work until the changes on the LSP server are merged?

tonklon:
I don't see any advantage holding this back.

This code behaves wrong when run in a path with spaces and this PR tries to fix this. If the fix makes sense and improves the output of this project, it should be merged IMHO. There is no interdependency with the fixes in the server.

vinistock:
FYI, we'll need to move the `BUNDLE_GEMFILE` part since not every OS handles env variable the same way #712.

One question: do the spaces need to be escaped inside of the string or is wrapping it around quotes enough? That is, do we need `/something\ other/` instead of `/something other`?

tonklon:
This PR is obsolete with #712. The solution to move the `BUNDLER_GEMFILE` into the environment works as well.

PR #717, opened 2023-07-24
WIP: Add command to re-run last test

andyw8:
### Motivation

Closes https://github.com/Shopify/ruby-dev-exp-issues/issues/1004

### DIFF
--- a/src/testController.ts
+++ b/src/testController.ts
@@ -22,6 +22,9 @@ export class TestController {
   private terminal: vscode.Terminal | undefined;
   private ruby: Ruby;
   private telemetry: Telemetry;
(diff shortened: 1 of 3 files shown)


src/testController.ts
```diff
@@ -150,6 +157,23 @@ export class TestController {
     });
   }
 
+  private async rerunLastTest() {
+    if (!this.lastTest) {
+      return;
+    }
+
+    // TODO: Add telemetry for this?
```
andyw8, line 165: If so, do we want to send a "rerun_last_event" in addition to the existing `"test"` or `"test_in_terminal"` command?

PR #721, opened 2023-07-25
Avoid running bundler commands before running custom bundle logic

vinistock:
### Motivation

I believe these are the issues reported here https://github.com/Shopify/vscode-ruby-lsp/pull/664#issuecomment-1646375400.

We can't make any bundler inquiries before starting the server. Now that the custom bundle logic runs on the server, the bundle will only be installed after the server booted successfully. Trying to read the server version before starting will always fail if the gems aren't installed (because we haven't run bundle install yet).

Additionally, we need to fully deprecate Ruby 2.7. Because the gem now requires Ruby >=3.0, using `gem install ruby-lsp` will never install `v0.7.3`, which means it's impossible to use the extension with Ruby 2.7.

### Implementation

- Moved the server version check after the custom bundle logic
- Changed the Ruby version check

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -211,9 +211,11 @@ export default class Client implements ClientInterface {
       }
     });
 
-    this.telemetry.serverVersion = await this.getServerVersion();
     await this.client.start();
+
+    // We cannot inquire anything related to the bundle before the custom bundle logic in the server runs
     await this.determineFormatter();
+    this.telemetry.serverVersion = await this.getServerVersion();
 
     this.state = ServerState.Running;
   }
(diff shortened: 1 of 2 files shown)


st0012 (approved):
Unfortunately I can't reproduce the issue I saw anymore, so I can't confirm that it'd be fixed with the change. But I think it's likely that the issue was caused by the mentioned problem 👍

bitwise-aiden (approved):

andyw8:
> Additionally, we need to fully deprecate Ruby 2.7

We should make sure to highlight this in the release notes.

vinistock:
Yes. This is a complete drop of 2.7.

PR #735, opened 2023-08-04
Add early return when shadowenv returns an empty string

bitwise-aiden:
### Motivation

Raised by @EsterKais, shadowenv has been failing because we are expecting `shadowenv hook --json` to return valid json. It is possible that this command won't print anything to stdout which will result in an empty string instead.

### Implementation

Check command output, returning early if it is empty.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -122,8 +122,11 @@ export class Ruby {
       cwd: this.workingFolder,
     });
 
-    if (result.stdout === "") return;
-
+    if (result.stdout === "") {
+      // eslint-disable-next-line no-process-env
+      this._env = { ...process.env };
+      return;
+    };
     // eslint-disable-next-line no-process-env
     const env = { ...process.env, ...JSON.parse(result.stdout).exported };
 
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -122,6 +122,8 @@ export class Ruby {
       cwd: this.workingFolder,
     });
 
+    if (result.stdout === "") return;
+
     // eslint-disable-next-line no-process-env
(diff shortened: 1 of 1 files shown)


src/ruby.ts
```diff
@@ -122,6 +122,11 @@ export class Ruby {
       cwd: this.workingFolder,
     });
 
+    if (result.stdout === "") {
+      // eslint-disable-next-line no-process-env
+      this._env = { ...process.env };
+      return;
+    };
```
paracycle, line 129: Alternatively:
```suggestion
    if (result.stdout === "") {
      result.stdout = "{ }";
    };
```

paracycle:
```js
> { ...{ foo: 1, bar: 2}, ...JSON.parse("{ }").exported }
{ foo: 1, bar: 2 }
```

src/ruby.ts
```diff
@@ -122,6 +122,11 @@ export class Ruby {
       cwd: this.workingFolder,
     });
 
+    if (result.stdout === "") {
```
paracycle, line 125: Could there be any whitespace in `stdout`? If so, we would be safer with a [`trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim):
```suggestion
    if (result.stdout.trim() === "") {
```

src/ruby.ts
vinistock: That's a neat idea.

src/ruby.ts
vinistock: I don't think so, but it doesn't hurt either.

EsterKais (approved):

PR #757, opened 2023-08-15
Narrow document selector to file and untitled schemes

vinistock:
### Motivation

In #745, we made the document selector a little too broad. There are some schemes that we're not interested in handling.

For example, when looking at a diff in VS Code, it opens both a `file` version of the document and a `git` version. Because both have the same path and VS Code sends requests for the two versions, they end up overriding each other and leads to weird issues.

### Implementation

Narrowed our selector to handle only file and untitled.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -87,7 +87,10 @@ export default class Client implements ClientInterface {
 
     const configuration = vscode.workspace.getConfiguration("rubyLsp");
     const clientOptions: LanguageClientOptions = {
-      documentSelector: [{ language: "ruby" }],
+      documentSelector: [
+        { scheme: "file", language: "ruby" },
+        { scheme: "untitled", language: "ruby" },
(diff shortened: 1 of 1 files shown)


vinistock:
Sorry, nevermind this PR. I think the fix lies elsewhere.

src/client.ts
```diff
@@ -87,7 +87,10 @@ export default class Client implements ClientInterface {
 
     const configuration = vscode.workspace.getConfiguration("rubyLsp");
     const clientOptions: LanguageClientOptions = {
-      documentSelector: [{ language: "ruby" }],
+      documentSelector: [
+        { scheme: "file", language: "ruby" },
+        { scheme: "untitled", language: "ruby" },
```
andyw8, line 92: I'm curious why the scheme isn't `git` ?

PR #786, opened 2023-09-13
Allow configuring when to pull diagnostics

vinistock:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/963

I just realized this is significantly easier to implement than I originally thought. The client's now have the option to decide when to pull diagnostics: on change, on save or both.

That makes it super easy to make it configurable.

### Implementation

1. Added a new configuration to allow devs to decide when to pull diagnostics
2. Started passing `diagnosticPullOptions` to the client initialization

### Manual tests

2. Configure `"rubyLsp.pullDiagnosticsOn": "save"`
3. Verify changing a file doesn't show RuboCop diagnostics
4. Verify saving the file still fixes violations
5. If you add a violation that can't be auto-correct (like a super long line), verify that you still see diagnostics after save

### DIFF
--- a/package.json
+++ b/package.json
@@ -249,6 +249,16 @@
           "description": "Run the Ruby LSP server from the specified branch rather than using the released gem. Only supported if not using bundleGemfile",
           "type": "string",
           "default": ""
+        },
+        "rubyLsp.pullDiagnosticsOn": {
+          "description": "When to pull diagnostics from the server (on change, save or both). Selecting save may significantly improve performance on large files",
+          "type": "string",
+          "enum": [
+            "change",
+            "save",
+            "both"
+          ],
+          "default": "both"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -251,7 +251,7 @@
           "default": ""
         },
         "rubyLsp.pullDiagnosticsOn": {
-          "description": "When to pull diagnostics from the server (on change, save or both). Selecting save may significantly improve performance on large files",
+          "description": "When to pull diagnostics from the server (on change, save or both). Selecting 'save' may significantly improve performance on large files",
           "type": "string",
           "enum": [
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -587,4 +589,15 @@ export default class Client implements ClientInterface {
   private hasUserDefinedCustomBundle(): boolean {
     return this.customBundleGemfile.length > 0;
   }
+
+  private diagnosticPullOptions(): DiagnosticPullOptions {
+    const configuration = vscode.workspace.getConfiguration("rubyLsp");
+    const pullOn: "change" | "save" | "both" =
```
andyw8, line 595: Is there are a chance another option could be added in future, in which case `both` would no longer make sense? If so, could this be instead be named `change_or_save`, or similar?

vinistock: If you open a new file, diagnostics are always computed once in the beginning. Until the LSP clears the diagnostics for that file, they never go away - even if you switch files.

I think we can go with `both` and change in the future if it ever becomes necessary.

package.json
```diff
@@ -249,6 +249,16 @@
           "description": "Run the Ruby LSP server from the specified branch rather than using the released gem. Only supported if not using bundleGemfile",
           "type": "string",
           "default": ""
+        },
+        "rubyLsp.pullDiagnosticsOn": {
+          "description": "When to pull diagnostics from the server (on change, save or both). Selecting save may significantly improve performance on large files",
```
andyw8, line 254:
```suggestion
          "description": "When to pull diagnostics from the server (on change, save or both). Selecting 'save' may significantly improve performance on large files",
```

andyw8 (approved):

bitwise-aiden (approved):

Issue #781, opened 2023-09-08
Review test running associated commands


vinistock:
We added specific commands to be able to run tests, but after double-checking the default keybindings I realized that this is already possible out of the box.

There are several commands to run or debug tests, like
- testing.debugAll
- testing.runAll
- testing.debugAtCursor
- testing.runCurrentFile
- testing.runAtCursor
- testing.reRunLastRun
- testing.reRunFailTests

I don't think we needed the implementation from #696 and I don't believe we need the implementation from #708 and #717.

Let's confirm that this is the case and
- Revert the changes in #696 
- Document what the default keybindings for interacting with the test explorer are somewhere. People can easily change their keybindings if they want to

andyw8:
I verified these work when called from the command menu:

- testing.runAtCursor
- testing.debugAtCursor
- testing.runCurrentFile
- testing.reRunFailTests (odd that it's `fail` and not `failed`)
- testing.reRunLastRun

`testing.runAll` also works, but for our extension I don't think there would be any difference than `testing.runCurrentFile`, since the explorer only covers the currently active file.

Something else I noticed - many of these already have keybinding. They use Command-; like a modifier, e.g. to re-run the last test it's Command-; followed by l.

The one that doesn't work currently is `testing.debugAll`.

PR #797, opened 2023-09-19
Remove duplicate test runner commands, document keybindings

andyw8:
### Motivation

Closes https://github.com/Shopify/vscode-ruby-lsp/issues/781, #717. Partially reverts #696.

### Implementation

Remove the redundant commands. Document the default keybindings.

I've kept the 'Run in Terminal' command for now until we decide what to about that.

### Automated Tests

None since we are using VS Code's built-in behaviour.

### Manual Tests

Run in debug mode and check against Ruby LSP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,24 @@ Possible values are:
 - `messages`: display requests and responses notifications
 - `verbose`: display each request and response as JSON
 
+### Keybindings
+
+VS Code provides the default keybindings:
+
+| Command                        | Mac                 | Windows/Linux       |
+| ------------------------------ | ------------------- | ------------------- |
+| Test: Run Test At Cursor       | Command-; C         | Control-; C         |
+| Test: Rerun Last Run           | Command-; L         | Control-; L         |
+| Test: Rerun Failed Tests       | Command-; E         | Control-; E         |
+| Test: Run Test In Current File | Command-; F         | Control-; F         |
+| Test: Debug Test At Cursor     | Command-; Command-C | Control-; Control-C |
+
+In addition, the extension provides keybindings for:
+
+| Command                                | Mac           | Windows/Linux |
+| -------------------------------------- | ------------- | ------------- |
+| Ruby LSP: Run Current Test In Terminal | Ctrl-Option-I | Control-Alt-I |
+
 ### Debugging the server using VS Code
 
 The `launch.json` contains a 'Minitest - current file' configuration for the debugger.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -286,6 +286,15 @@ Possible values are:
 - `messages`: display requests and responses notifications
 - `verbose`: display each request and response as JSON
 
+### Default Keybindings
+
+|Command|Mac|Windows/Linux|
+|-|-|-|
+|Test: Run Test At Cursor|Command-; C|Control-; C|
```
andyw8, line 293: TODO: Confirm the Windows/Linux ones are actually correct.

src/testController.ts
```diff
@@ -70,20 +70,10 @@ export class TestController {
 
     context.subscriptions.push(
       this.testController,
-      vscode.commands.registerCommand(
```
vinistock, line 73: These commands are used by the code lens, aren't they?

vinistock (changes requested):

src/testController.ts
andyw8: Yes, that was a mistake, now restored. I had tested it but must not have realised that the extension host environment wasn't active.

README.md
```diff
@@ -267,6 +267,24 @@ Possible values are:
...
+| Test: Rerun Last Run           | Command-; L         | Control-; L         |
+| Test: Rerun Failed Tests       | Command-; E         | Control-; E         |
+| Test: Run Test In Current File | Command-; F         | Control-; F         |
+| Test: Debug Test At Cursor     | Command-; Command-C | Control-; Control-C |
+
+In addition, the extension provides keybindings for:
+
+| Command                                | Mac           | Windows/Linux |
+| -------------------------------------- | ------------- | ------------- |
+| Ruby LSP: Run Current Test In Terminal | Ctrl-Option-I | Control-Alt-I |
```
andyw8, line 286: This is the mapping that dev-test-runner currently use, but it might be more cohesive to use something like `Command-; T`.

package.json
```diff
@@ -64,30 +64,26 @@
...
         "command": "rubyLsp.showSyntaxTree",
         "title": "Show syntax tree",
         "category": "Ruby LSP"
       }
     ],
+    "keybindings": [
+      {
+        "command": "rubyLsp.runTestInTerminal",
+        "key": "ctrl+alt+i",
+        "mac": "ctrl+alt+i",
```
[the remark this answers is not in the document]
andyw8: I had taken it from `dev-test-runner` which uses `ctrl`, even on Mac. But as per other comment, this will change anyway.

PR #798, opened 2023-09-21
Use general middleware to report performance and crash telemetry

vinistock:
### Motivation

Closes Shopify/vscode-ruby-lsp#571, Closes Shopify/ruby-lsp#1498

A new version of the LSP client package has been released including our general middleware contribution. We can finally move forward with [removing our server telemetry](https://github.com/Shopify/ruby-lsp/pull/650) and shifting to client only.

### Implementation

- Upgraded the package to v9.0.0
- The new package requires the types for InlineCompletion, which only exist in VS Code >= 1.68. I updated our engine requirement and the types package
- Added the general middleware for both requests and notifications

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2,6 +2,7 @@ import path from "path";
 import fs from "fs";
 import { promisify } from "util";
 import { exec } from "child_process";
+import { performance as Perf } from "perf_hooks";
 
 import * as vscode from "vscode";
 import {
@@ -14,9 +15,10 @@ import {
   ExecutableOptions,
   ServerOptions,
   DiagnosticPullOptions,
+  MessageSignature,
 } from "vscode-languageclient/node";
 
-import { Telemetry } from "./telemetry";
+import { Telemetry, RequestEvent } from "./telemetry";
 import { Ruby } from "./ruby";
 import { StatusItems, Command, ServerState, ClientInterface } from "./status";
 import { TestController } from "./testController";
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -174,6 +176,29 @@ export default class Client implements ClientInterface {
...
+            type: string | MessageSignature,
+            param?: TP,
+            token?: vscode.CancellationToken,
+          ) => Promise<T>,
+        ) => {
+          return this.benchmarkMiddleware(type, param, () =>
+            next(type, param, token),
+          );
+        },
+        sendNotification: async <TR>(
```
andyw8, line 193: What does the `R` in `TR` mean here?

src/client.ts
```diff
@@ -600,4 +605,86 @@ export default class Client implements ClientInterface {
...
+        telemetryData.uri = uri;
+      }
+
+      telemetryData.params = JSON.stringify(params);
+    }
+
+    let result: T | undefined;
+    let errorResult;
+
+    const benchmarkId = Math.floor(Math.random() * 1000);
```
andyw8, line 649: Is this 'random enough'? Seems we'd often have duplicates.

src/client.ts
```diff
@@ -600,4 +605,86 @@ export default class Client implements ClientInterface {
...
+
+    // Execute the request measuring the time it takes to receive the response
+    Perf.mark(`${benchmarkId}.start`);
+    try {
+      // result = await next(type, param, token);
```
andyw8, line 654: ✂️

src/client.ts
vinistock: This ID is just for marking the start and end of the benchmark. I switched to use an incrementing ID instead to avoid having to do the calculations.

PR #800, opened 2023-09-21
Group minor and patch upgrades

vinistock:
### Motivation

Similar to what we did in the Ruby LSP, let's group minor and patch upgrades.


vinistock:
This is the VS Code extension repo. I don't think it'd be impacted by the Ruby dependencies.

PR #807, opened 2023-09-26
Set typescript module option to nodeNext

andyw8:
### Motivation

As per the suggestion in https://github.com/isaacs/node-glob/issues/560#issuecomment-1735868693, to solve the problem we had with the latest `glob` update.

I'm not sure if there are any potential downsides of this approach, maybe someone with more Js/TS knowledge can say.

Also closes https://github.com/Shopify/vscode-ruby-lsp/issues/806

### Implementation

Update the tsconfig.

### Automated Tests

n/a

### Manual Tests

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -437,7 +437,7 @@
     "eslint": "^8.50.0",
     "eslint-import-resolver-typescript": "^3.6.1",
     "eslint-plugin-prettier": "^5.0.0",
-    "glob": "^10.3.7, < 10.3.7",
+    "glob": "^10.3.7",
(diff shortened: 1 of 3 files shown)


vinistock:
The change looks good to me, but let's please confirm two things:

1. Is the extension working normally with this new configuration?
2. Can we check what extensions that are bundled with VS Code itself are doing? If at least one of them uses that option, then it's probably safe to say we can do it too

andyw8:
No longer need with the latest `glob` release.

PR #824, opened 2023-10-11
Add support for HAML syntax highlighting in Heredoc sections

joevandyk:
### Motivation

We have support for C++, JS, SQL, HTML, etc syntax highlighting in Heredoc, would be good to have it for HAML as well. Main use case for me is embedding HAML templates inline in ViewComponents. 

### Implementation

Copy/Pasted the HTML section.

### Automated Tests

None, the other Heredoc syntax highlighting don't seem to have any.

### Manual Tests
image

### DIFF
--- a/grammars/ruby.cson.json
+++ b/grammars/ruby.cson.json
@@ -1557,6 +1557,44 @@
       ]
     },
     {
+      "begin": "(?=(?><<[-~]([\"'`]?)((?:[_\\w]+_|)HAML)\\b\\1))",
+      "comment": "Heredoc with embedded HAML",
+      "end": "(?!\\G)",
+      "name": "meta.embedded.block.haml",
+      "patterns": [
+        {
+          "begin": "(?><<[-~]([\"'`]?)((?:[_\\w]+_|)HAML)\\b\\1)",
+          "beginCaptures": {
+            "0": {
+              "name": "punctuation.definition.string.begin.ruby"
+            }
+          },
+          "contentName": "text.haml",
+          "end": "^\\s*\\2$\\n?",
+          "endCaptures": {
+            "0": {
+              "name": "punctuation.definition.string.end.ruby"
+            }
+          },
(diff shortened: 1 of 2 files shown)


paracycle:
I am not sure if we should keep adding highlighting for other languages inside Heredocs, since VS Code already does proper syntax highlighting for heredocs, provided a separate extension is installed for syntax highlighting the target language.

For example, compare the following two screenshots from VS Code:
image
image

Since I had the [GraphQL syntax highlighting extension](https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql-syntax) installed in my system, VS Code was able to highlight embedded GraphQL query in a heredoc properly, provided the heredoc label is `GRAPHQL`

joevandyk:
Doesn’t that only work because there’s an entry for heredoc graphql here? 

https://github.com/Shopify/vscode-ruby-lsp/blob/18476815807165ac84919ec5e7279ca3788a6a52/grammars/ruby.cson.json#L1679-L1715

davidstosik:
> Doesn’t that only work because there’s an entry for heredoc graphql here?

Yeah, I think so.
In a virgin VSCode with only the GraphQL extension, I'm not getting GraphQL syntax highlighting:
image

But I get HTML syntax highlighting, because VSCode's out-of-the-box Ruby syntax/grammar declares a "Heredoc with embedded html":

https://github.com/microsoft/vscode/blob/0c36b92c82064882a228487040187cfc13669c0f/extensions/ruby/syntaxes/ruby.tmLanguage.json#L590-L627

sambostock:
#842 is relevant to this PR. If it's merged first, we'll want to update the test config it adds to add `HAML` (see #843 for `YAML` example).

vinistock:
Please let us know if it's working as you expected and then we can merge.

joevandyk:
it looks good! 
image

vinistock (approved):

PR #825, opened 2023-10-11
Fix broken version manager link

vinistock:
### Motivation

This link doesn't work if we don't include the full URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,8 +101,8 @@ default shell.
 
 > [!NOTE]
 >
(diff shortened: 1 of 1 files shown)


andyw8 (approved):
I was confused why the relative link didn't work, since it does elsewhere, e.g. [here](https://github.com/Shopify/ruby-lsp/blob/main/README.md?plain=1#L23). Must be something to do with it being in a `>` section?

PR #833, opened 2023-10-13
Better explain Ruby version requirements

vinistock:
### Motivation

With some frequency, people with EOL Ruby versions ask about why the Ruby LSP is failing. I think it's worth better documenting the requirement and pointing to the documentation from the error dialogue so that it's easier to find alternatives.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Want to discuss Ruby developer experience? Consider joining the public
 
 Search for `Shopify.ruby-lsp` in the extensions tab and click install.
 
-By default, the Ruby LSP will generate a `.ruby-lsp` folder with a custom bundle that includes the server gem.
+By default, the Ruby LSP will generate a `.ruby-lsp` directory with a custom bundle that includes the server gem.
 Additionally, it will attempt to use available version managers to select the correct Ruby version for any given
 project. Refer to configuration for more options.
 
@@ -116,25 +116,65 @@ The tool to be used for formatting files can be configured with the following se
 "rubyLsp.formatter": "auto"
 ```
 
+#### Ruby version requirement
+
+By default, the Ruby LSP uses the current project's Ruby version and bundle. This allows the LSP to index the correct
+gem versions, and to ensure formatting behavior is consistent with CI.
+
+The Ruby LSP and its main dependency [Prism](https://github.com/ruby/prism) (the new Ruby parser) both follow the same
+policy, which is to support only Ruby versions that are not beyond their end-of-life.
+
+If you're working on a project with an older Ruby version, it might be possible to install older versions of the server
+gem to get support for older rubies, but that might also involve using older versions of the VS Code extension - since
+some functionality requires implementations in both client and server.
+
+The other alternative is to use a custom Gemfile separate from the project with a different Ruby version. Notice that
+certain functionality may be degraded or require manual configuration, since the Ruby LSP will not be able to inspect
+the project's real bundle to discover dependencies. Please see the instructions below.
+
 #### Using a custom Gemfile
 
 If you are working on a project using an older version of Ruby not supported by Ruby LSP, then you may specify a
-separate `Gemfile` for development tools. You can include the `ruby-lsp` in it and point to that Gemfile by using the
-following configuration:
+separate `Gemfile` for development tools.
 
 **Note**: when using this, gems will not be installed automatically and neither will `ruby-lsp` upgrades.
 
+Create a directory to store the custom bundle outside of the project that uses the old Ruby version. Inside that
+directory, add your preferred version manager configuration to select a supported Ruby version. For example, if using
+`chruby`, it would look like this:
+
+```shell
+# the/directory/.ruby-version
+
+3.2.2
+```
+
+Create a `Gemfile` for development tools inside that directory.
+
+```ruby
+# the/directory/Gemfile
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "ruby-lsp"
+gem "rubocop"
+```
+
+Run `bundle install` inside that directory to generate a lockfile. After the directory contains the custom `Gemfile` and
+the version manager configuration, use the following configuration in VS Code to point the Ruby LSP to that `Gemfile`.
+
 ```jsonc
 {
-  "rubyLsp.bundleGemfile": "../../relative/path/to/Gemfile", // using a relative path from the current project
-  "rubyLsp.bundleGemfile": "/absolute/path/to/Gemfile" // using an absolute path
+  "rubyLsp.bundleGemfile": "../../relative/path/to/the/directory/Gemfile",
+  "rubyLsp.bundleGemfile": "/absolute/path/to/the/directory/Gemfile"
 }
 ```
 
 #### Configuring VS Code debugger
 
-To configure the VS Code debugger, you can use the "Debug: Add configuration..." command to create a
-`launch.json` file in the `.vscode` folder of your project.
+To configure the VS Code debugger, you can use the "Debug: Add configuration..." command to create a `launch.json` file
+in the `.vscode` directory of your project.
 
 This command would generate the following configuration:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -116,18 +116,59 @@ The tool to be used for formatting files can be configured with the following se
 "rubyLsp.formatter": "auto"
 ```
 
+#### Ruby version requirement
+
+By default, the Ruby LSP always uses the current project's Ruby version and bundle. This allows the LSP to find the
```
andyw8, line 121:
```suggestion
By default, the Ruby LSP uses the current project's Ruby version and bundle. This allows the LSP to find the
```

README.md
```diff
@@ -116,18 +116,59 @@ The tool to be used for formatting files can be configured with the following se
...
+correct gem versions automatically to index and to format the code matching behavior on CI.
+
+The Ruby LSP and its main dependency Prism (the new Ruby parser) both follow the same policy, which is to support only
```
andyw8, line 124:
```suggestion
The Ruby LSP and its main dependency [Prism](https://github.com/ruby/prism) (the new Ruby parser) both follow the same policy, which is to support only
```

README.md
```diff
@@ -116,18 +116,59 @@ The tool to be used for formatting files can be configured with the following se
 "rubyLsp.formatter": "auto"
 ```
 
+#### Ruby version requirement
+
+By default, the Ruby LSP always uses the current project's Ruby version and bundle. This allows the LSP to find the
+correct gem versions automatically to index and to format the code matching behavior on CI.
```
andyw8, line 122:
```suggestion
correct gem versions automatically so that index and formatting behaviour is consistent across all development and test environments.
```

README.md
```diff
@@ -116,18 +116,59 @@ The tool to be used for formatting files can be configured with the following se
...
+
+The Ruby LSP and its main dependency Prism (the new Ruby parser) both follow the same policy, which is to support only
+Ruby versions that are not beyond their end-of-life.
+
+If you're working on a project with an older Ruby version, it might be possible to install older versions of the server
+gem to get support for older rubies, but that might also involve using older versions of the VS Code extension - since
+some functionality requires implementations in both client and server.
+
+The other alternative is to use a custom Gemfile separate from the project with a different Ruby version. Notice that
```
andyw8, line 131:
```suggestion
The alternative is to use a custom Gemfile separate from the project with a different Ruby version. Notice that
```

README.md
```diff
@@ -116,18 +116,59 @@ The tool to be used for formatting files can be configured with the following se
...
+certain functionality may be degraded or require manual configuration, since the Ruby LSP will not be able to inspect
+the project's real bundle to discover dependencies. Please, see the instructions below.
```
andyw8, line 133:
```suggestion
the project's real bundle to discover dependencies. Please see the instructions below.
```

README.md
```diff
@@ -116,18 +116,59 @@ The tool to be used for formatting files can be configured with the following se
...
 
 If you are working on a project using an older version of Ruby not supported by Ruby LSP, then you may specify a
-separate `Gemfile` for development tools. You can include the `ruby-lsp` in it and point to that Gemfile by using the
-following configuration:
+separate `Gemfile` for development tools.
 
 **Note**: when using this, gems will not be installed automatically and neither will `ruby-lsp` upgrades.
 
+Create a folder to store the custom bundle outside of the project that uses the old Ruby version. Inside that folder,
+add your favorite version manager configuration to select a supported Ruby version. For example, if using `chruby`, it
```
andyw8, line 143:
```suggestion
add your preferred version manager configuration to select a supported Ruby version. For example, if using `chruby`, it
```

README.md
```diff
@@ -116,18 +116,59 @@ The tool to be used for formatting files can be configured with the following se
...
 #### Using a custom Gemfile
 
 If you are working on a project using an older version of Ruby not supported by Ruby LSP, then you may specify a
-separate `Gemfile` for development tools. You can include the `ruby-lsp` in it and point to that Gemfile by using the
-following configuration:
+separate `Gemfile` for development tools.
 
 **Note**: when using this, gems will not be installed automatically and neither will `ruby-lsp` upgrades.
 
+Create a folder to store the custom bundle outside of the project that uses the old Ruby version. Inside that folder,
```
andyw8, line 142: (nit) we should really say `directory` rather than `folder`, since folder is a GUI abstraction.

README.md
```diff
@@ -116,18 +116,59 @@ The tool to be used for formatting files can be configured with the following se
...
+# the/folder/Gemfile
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "ruby-lsp"
+gem "rubocop"
+```
+
+Don't forget to run `bundle install` inside that folder to generate a lockfile. After the folder contains the custom
```
andyw8, line 164:
```suggestion
Run `bundle install` inside that folder to generate a lockfile. After the folder contains the custom
```

src/ruby.ts
```diff
@@ -192,7 +192,8 @@ export class Ruby {
 
     if (major < 3) {
       throw new Error(
-        "The Ruby LSP requires Ruby 3.0 or newer to run. Please upgrade your Ruby version",
+        "The Ruby LSP requires Ruby 3.0 or newer to run. \
```
andyw8, line 195:
```suggestion
        "The Ruby LSP officially supports only Ruby 3.0 or newer. \
```

src/ruby.ts
```diff
@@ -192,7 +192,8 @@ export class Ruby {
...
+        [Learn more](https://github.com/Shopify/vscode-ruby-lsp?tab=readme-ov-file#ruby-version-requirement)",
```
andyw8, line 196:
```suggestion
        [Learn about using with older Ruby versions](https://github.com/Shopify/vscode-ruby-lsp?tab=readme-ov-file#ruby-version-requirement)",
```

vinistock: Do you think we need to be this explicit given that the message already mentions the Ruby version?

src/ruby.ts
vinistock: I think I still prefer the `requires` here instead of `officially supports`. To me it sounds like there's unofficial support for older rubies, which isn't the case.

src/ruby.ts
andyw8: My concern is that if people see "The Ruby LSP requires Ruby 3.0 or newer to run" then it sounds absolute, so why would they click to learn more.

README.md
vinistock: I don't think this says exactly what I meant. Indexing doesn't happen in CI, it's just formatting that needs to match it.

README.md
andyw8: Ah I see, ignore.

README.md
andyw8: Ok, how about:

> By default, the Ruby LSP uses the current project's Ruby version and bundle. This allows the LSP to index the correct gem versions, and to ensure formatting behaviour is consistent across all development and test environments.

src/ruby.ts
andyw8: Fair, but I think there still needs to be some subtlety in there to indicate that there is a way to work with older Rubies.

src/ruby.ts
vinistock: I changed to `See alternatives`.

src/ruby.ts
```diff
@@ -188,11 +188,12 @@ export class Ruby {
     this.rubyVersion = rubyVersion;
     this.yjitEnabled = yjitIsDefined === "constant";
 
-    const [major, minor, _patch] = this.rubyVersion.split(".").map(Number);
+    const [major, minor, patch] = this.rubyVersion.split(".").map(Number);
 
     if (major < 3) {
       throw new Error(
-        "The Ruby LSP requires Ruby 3.0 or newer to run. Please upgrade your Ruby version",
+        `The Ruby LSP requires Ruby 3.0 or newer to run. This project is using ${major}.${minor}.${patch}. \
```
andyw8, line 195:
```suggestion
        `The Ruby LSP requires Ruby 3.0 or newer to run. This project is using ${this.rubyVersion}. \
```

andyw8 (approved):

Issue #832, opened 2023-10-13
Debugging rails app


stage-rl:
### Operating System

Debian Linux

### Ruby version

3.2.2

### Project has a bundle

- [X] Has bundle

### Ruby version manager being used

rbenv

### Description

Hello, tried to use the debugger for rails apps in "attach mode" and gave up several times returning to [vscode-rdbg](https://github.com/ruby/vscode-rdbg)

Only after seeing [this](https://github.com/Shopify/vscode-ruby-lsp/issues/829#issuecomment-1759862059) comment I tried running my rails app as `bundle exec rdbg --open --command --sock-path=/tmp/ruby-lsp-debug-sockets/test.sock -- bin/rails server -p 3000` and it works. This still required to manually create /tmp/ruby-lsp-debug-sockets folder.  Still this is probably not the required approach.

Can you pls add info on proper procedure to README? Thanks a lot and keep up the good work

vinistock:
Thank you for the bug report! I'm puzzled as to how it worked without any manual intervention in `vscode-rdbg`.

The only way to attach to an existing Ruby process is if that process was started with the debugger in the first place. Do you have any `rdbg` configuration?

When you attach with `rdbg`, how do you launch the server?

stage-rl:
With `vscode-rdbg` it was slightly simpler `rdbg -O -n -c -- bin/rails server -p 3000`

vinistock:
The idea of the socket path is that we create those automatically for you when launching the debugger, so that you don't have to configure or select anything.

For attach, it's a bit different though. The extension has no way of knowing where the socket was created, so that it can offer as an option for you to select.

In terms of the command, I'm pretty they are equivalent, we're just using the long names for the arguments (i.e.: `--open` instead of `-O`).

The Ruby LSP debug client will default to look for sockets inside `/tmp/ruby-lsp-debug-sockets`.

@st0012, in retrospect that was a bit of an odd decision from our part. If you don't specify a `sock-path`, does Ruby debug always put the sockets in the same folder? Maybe we should change this to use whatever it does by default.

vinistock:
FWIW, it's already possible to launch a Rails server with the debugger using the configuration Stan mentioned. But in this attach scenario, it's definitely not clear.

I think we should

[...]

PR #846, opened 2023-10-18
Add more debugging logs

vinistock:
### Motivation

Closes #829
Improves #832 a bit

As more people begin to use the extension, we're subject to a new variety of configurations and environments and it's often hard to tell what's going without detailed information. I think we need to log more in the extension, so that debugging complex setups is easier.

### Implementation

Started sharing the output channel between our main objects and then printed a bunch of logs to it. It should help diagnose issues more easily.

Are there any other things we should be logging?

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -16,13 +16,16 @@ export class Debugger
   private debugProcess?: ChildProcessWithoutNullStreams;
   private readonly console = vscode.debug.activeDebugConsole;
   private readonly subscriptions: vscode.Disposable[];
+  private readonly outputChannel: vscode.OutputChannel;
 
   constructor(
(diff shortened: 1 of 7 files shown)


src/debugger.ts
```diff
@@ -214,8 +229,9 @@ export class Debugger
       // actually an error
       this.debugProcess.on("exit", (code) => {
         if (code) {
-          const message = `Debugger exited with status ${code}`;
+          const message = `Debugger exited with status ${code}. Check the output channel for more information`;
```
andyw8, line 232:
```suggestion
          const message = `Debugger exited with status ${code}. Check the output channel for more information.`;
```

andyw8 (approved):

PR #851, opened 2023-10-19
Fix alternative Ruby activation

### LINKED ISSUE #849: Trying to work with a custom gemfile doesn`t work
### Operating System

MacOS

### Ruby version

3.2.2

### Project has a bundle

- [X] Has bundle

chruby

### Description

I'm trying to set up RUBY LSP to run on VSCode. We have a project that is currently running in Docker with version 2.7.7, and even following the tutorial in the README, it's not working.

I tried creating the folder with version 3.2.2, but whenever I try to run the project, the following error is given: "Failed to activate chruby environment: The Ruby LSP requires Ruby 3.0 or newer to run. Please upgrade your Ruby version."

It seems like it's completely ignoring the configuration in settings:

"rubyLsp.bundleGemfile": "/Users/lucasluan/Documents/Work/utils/Gemfile"

vinistock:
### Motivation

Closes #849

When someone specifies a custom bundle gemfile in order to use a Ruby version different than their project, we need to read the `.ruby-version` file from the directory where the custom bundle exists.

Additionally, we need to use the correct directory when fetching Ruby information or else we end up activating the correct Ruby version, but displaying the wrong one in the status item.

### Implementation

Put the `cwd` in a property and started using it everywhere we need it.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -30,6 +30,7 @@ export class Ruby {
   private _error = false;
   private readonly context: vscode.ExtensionContext;
   private readonly customBundleGemfile?: string;
+  private readonly cwd: string;
   private readonly outputChannel: vscode.OutputChannel;
 
   constructor(
@@ -50,6 +51,10 @@ export class Ruby {
         ? customBundleGemfile
         : path.resolve(path.join(this.workingFolder, customBundleGemfile));
     }
+
+    this.cwd = this.customBundleGemfile
+      ? path.dirname(this.customBundleGemfile)
+      : this.workingFolder;
   }
 
   get versionManager() {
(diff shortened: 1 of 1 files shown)


src/ruby.ts
```diff
@@ -248,7 +242,9 @@ export class Ruby {
   }
 
   private readRubyVersion() {
-    let dir = this.workingFolder;
+    let dir = this.customBundleGemfile
+      ? path.dirname(this.customBundleGemfile)
+      : this.workingFolder;
```
paracycle, line 247: I am confused about this one. Why can't this be just:
```suggestion
    let dir = this.cwd;
```

PR #853, opened 2023-10-20
Pass experimental flag as CLI argument to the server

vinistock:
### Motivation

Client side of https://github.com/Shopify/ruby-lsp/pull/1126.

Get rid of the branch configuration and use `experimental` instead to install beta versions of the server gem.

### Implementation

Started passing the `--experimental` flag to the server.

### DIFF
--- a/package.json
+++ b/package.json
@@ -251,11 +251,6 @@
           "type": "integer",
           "default": 30
         },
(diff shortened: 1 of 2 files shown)


vinistock:
Closing this for now. When we switch to Bundler compose, this will change significantly.

Issue #850, opened 2023-10-19
Header must provide a Content-Length property


jcagarcia:
### Operating System

OSX Sonoma 14.0

### Ruby version

3.2.2

### Project has a bundle

- [X] Has bundle

### Ruby version manager being used

rvm

### Description

After opening a Ruby project using VSCode, I'm getting the following error when trying to initialize the Ruby LSP extension

```
Ruby LSP> Skipping custom bundle setup since /Users/Juancarlos.Garcia/code/my-project/.ruby-lsp/Gemfile.lock already exists and is up to date
Ruby LSP> Running bundle install for the custom bundle. This may take a while...
The Gemfile's dependencies are satisfied
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies...
Bundler attempted to update ruby-lsp but its version stayed the same
Bundler attempted to update debug but its version stayed the same
Bundle updated!
Starting Ruby LSP...
[Error - 20:47:53] Client Ruby LSP: connection to server is erroring.
Header must provide a Content-Length property.
{"resolving dependencies...\ncontent-length":"1539"}
Shutting down server.
[Error - 20:47:53] Stopping server failed
Error: Client is not running and can't be stopped. It's current state is: starting
	at $d.shutdown (/Users/Juancarlos.Garcia/.vscode/extensions/shopify.ruby-lsp-0.4.10/out/extension.js:41:8468)
	at $d.stop (/Users/Juancarlos.Garcia/.vscode/extensions/shopify.ruby-lsp-0.4.10/out/extension.js:41:8047)
	at $d.stop (/Users/Juancarlos.Garcia/.vscode/extensions/shopify.ruby-lsp-0.4.10/out/extension.js:44:38188)
	at $d.handleConnectionError (/Users/Juancarlos.Garcia/.vscode/extensions/shopify.ruby-lsp-0.4.10/out/extension.js:43:63)
	at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[Error - 20:47:53] Client Ruby LSP: connection to server is erroring.
Header must provide a Content-Length property.
[...]

PR #868, opened 2023-11-01
Avoid printing to stdout from Client or Ruby activation

vinistock:
### Motivation

This may be the solution for #850.

I suspect the issue is that RVM is somehow causing something to be printed to STDOUT, which is basically left in the pipe until the LSP starts. When it finally does start, it reads STDOUT and gets the stuff that was printed, which is invalid from the LSP's perspective.

I'm not 100% sure this is the fix, but in general we should simply never use `STDOUT` or `STDIN` for anything when it comes to the client, because it may mess up the client-server communication. 

### Implementation

I switched all places in `Client` and `Ruby` where we were using STDOUT to get versions or perform activation. We now rely only on STDERR, which in LSPs is only used for printing and not for communication.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -432,13 +432,13 @@ export default class Client implements ClientInterface {
       "rubyLsp.lastGemUpdate",
     );
 
-    const { stdout } = await asyncExec("gem list ruby-lsp", {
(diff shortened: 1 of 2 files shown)


andyw8:
Might another approach be to 'flush' stdout before starting ruby-lsp?

andyw8 (approved):

PR #876, opened 2023-11-04
Stage rl/standard debug sockets

stage-rl:
### Motivation

Closes [#832](https://github.com/Shopify/vscode-ruby-lsp/issues/832)

Simplify debugging rails app and running tests. Use standard debug sockets provided by rdbg as much as possible.

### Implementation

Attach configuration - request sockets by running `bundle exec rdbg --util=list-socks`
Running tests - extract socket info from rdbg output
Updated Readme.md to reflect this

### Automated Tests

No tests added. If required, let me know

### Manual Tests

Install `debug` gem. Verify by running `bundle exec rdbg -v`
Run your app with rdbg - e.g. `bundle exec rdbg -O -n -c -- bin/rails server -p 3000`
You should be able to use vscode debugger attach configuration to connect to your running app


stage-rl:
I signed the CLA, check will hopefully succeed after review and changes

README.md
```diff
@@ -51,12 +51,13 @@ documentation](https://shopify.github.io/ruby-lsp/RubyLsp/Requests.html).
...
-| Ruby LSP: Restart                    | Restart the Ruby LSP server                             |
-| Ruby LSP: Stop                       | Stop the Ruby LSP server                                |
-| Ruby LSP: Update language server gem | Updates the `ruby-lsp` server gem to the latest version |
+
+| Command                              | Description                                            |
+| -------------------------------------- | -------------------------------------------------------- |
+| Ruby LSP: Start                      | Start the Ruby LSP server                              |
+| Ruby LSP: Restart                    | Restart the Ruby LSP server                            |
+| Ruby LSP: Stop                       | Stop the Ruby LSP server                               |
+| Ruby LSP: Update language server gem | Updates the`ruby-lsp` server gem to the latest version |
```
vinistock, line 60: What changed here? Can we remove this change?

README.md
```diff
@@ -203,6 +204,16 @@ This command would generate the following configuration:
 }
 ```
 
+### Debugging rails applications
```
vinistock, line 207:
```suggestion
### Debugging live processes
```

README.md
```diff
@@ -203,6 +204,16 @@ This command would generate the following configuration:
...
+
+Install `debug` gem. Verify by running `bundle exec rdbg -v`
+
+Make sure you're running your application with rdbg, e.g. `bundle exec rdbg -O -n -c -- bin/rails server -p 3000`
```
vinistock, line 211:
```suggestion
Run your application with the debugger attached, so that the extension can connect to it.

bundle exec rdbg -O -n -c -- bin/rails server -p 3000
```

README.md
```diff
@@ -203,6 +204,16 @@ This command would generate the following configuration:
...
+
+You should be able to use vscode debugger attach configuration to connect to your running app
```
vinistock, line 213: Launching the attach task should display a list of existing sockets for your running applications. Select the one you'd like to connect to and start debugging.

README.md
```diff
@@ -203,6 +204,16 @@ This command would generate the following configuration:
...
+
+For better integrated rails tests support also install `ruby-lsp-rails` gem
```
vinistock, line 215:
```suggestion
For better integrated rails tests support also install `ruby-lsp-rails` gem.
```

src/debugger.ts
```diff
@@ -121,41 +121,44 @@ export class Debugger
...
+    let sockets: string[] = [];
+    try {
+      sockets = execSync(cmd)
+        .toString()
+        .split("\n")
+        .filter((socket) => socket.length > 0);
+    } catch (error: any) {
+      this.console.append(error.message);
+    }
+    return sockets;
```
vinistock, line 135: Let's import `asyncExec` from common instead of using synchronous exec.

```suggestion
    try {
      const result = asyncExec(cmd, { cwd: this.workingFolder, env: this.ruby.env });
      
      return result.stdout
        .toString()
        .split("\n")
        .filter((socket) => socket.length > 0);
    } catch (error: any) {
      this.console.append(`Error listing sockets: ${error.message}`);
    }
```

src/debugger.ts
```diff
@@ -121,41 +121,44 @@ export class Debugger
...
-              new vscode.DebugAdapterNamedPipeServer(
-                path.join(socketsDir, selectedSocket),
-              ),
-            );
-          }
-        });
+        reject(new Error(`No debuggee processes found.`));
+      } else if (sockets.length === 1)
+        resolve(new vscode.DebugAdapterNamedPipeServer(sockets[0]));
+      else
```
vinistock, line 149: Let's always use braces here. E.g.:

```typescript
if () {
} else if () {
} else {
}
```

src/debugger.ts
```diff
@@ -173,7 +175,6 @@ export class Debugger
         "rdbg",
         "--open",
         "--command",
-        `--sock-path=${sockPath}`,
```
vinistock, line 176: This is still necessary for launch. In launch operations, the extension defines where the socket should be placed, so that it knows how to auto-connect.

We only want to change attach.

src/debugger.ts
```diff
@@ -209,7 +210,14 @@ export class Debugger
...
           initialized = true;
-          resolve(new vscode.DebugAdapterNamedPipeServer(sockPath));
+          const regex =
+            /DEBUGGER: Debugger can attach via UNIX domain socket \((.*)\)/;
+          const sockPath = RegExp(regex).exec(initialMessage);
+          if (sockPath && sockPath.length === 2) {
+            resolve(new vscode.DebugAdapterNamedPipeServer(sockPath[1]));
+          } else {
+            reject(new Error("Debugger not found on UNIX socket"));
+          }
```
vinistock, line 220: Yeah, this is not necessary if we keep our standard socket path for launch.

yarn-error.log
vinistock, line 1: This should not be committed.

README.md
stage-rl: For people searching for Rails debugging I suggest keeping at least `Debugging live processes (e.g. Rails applications)`

README.md
stage-rl: Sure, no change intended, vscode Markdown editor got too aggresive

src/debugger.ts
stage-rl: I added the cwd and env and applied suggested error handling improvement, unfortunately I can't find asyncExec in common or anywhere else actually. Pls point me to what exactly you mean.

child_process contains asynchronous `exec`. This could be used, though we're waiting for the result anyway as it's to be used immediately. So I'm thinking there would be no real benefit from complicating the implementation.

The docs say:
The child_process.execSync() method is generally identical to [child_process.exec()](https://nodejs.org/docs/latest-v16.x/api/child_process.html#child_processexeccommand-options-callback) with the exception that the method will not return until the child process has fully closed.

src/debugger.ts
stage-rl: The point is that we maybe don't want to hardcode those sockets and want to let rdbg decide. We then extract them from debugger output. The implementation provided handles both ends, see your other comment. 

> Yeah, this is not necessary if we keep our standard socket path for launch.

README.md
vinistock: Agreed. Can we switch the title and instead add a phrase explaining what this is then? Something along the lines of

`Instead of launching a process to debug every time, you may want to attach the VS Code debugger to an existing process, such as a Rails server. Follow these instructions to do so.`

src/debugger.ts
vinistock: @st0012 I think this is fine, but wanted to double-check to see if you have any concerns.

src/debugger.ts
```diff
@@ -121,41 +120,45 @@ export class Debugger
     this.subscriptions.forEach((subscription) => subscription.dispose());
   }
 
+  private getSockets(): string[] {
```
st0012, line 123: When getting sockets, we should take users' `launch.json` configuration into account, especially the `env`. When users specifying things like `RUBY_DEBUG_SOCK_DIR` when starting the debuggee process, the `rdbg --util=list-socks` call needs to have the same value or it wouldn't find anything.

The `vscode-rdbg` extension [takes the configuration into account](https://github.com/ruby/vscode-rdbg/blob/master/src/extension.ts#L407C28-L412) so it supports the above use case.

vinistock (approved):

PR #886, opened 2023-11-08
Refactor extension and allow for multiple workspaces

### LINKED ISSUE #381: Support using the LSP with multiple projects in the same VS Code instance

vinistock:
### Motivation

Closes #381

This pull request is a massive refactor to decouple many concepts in our codebase. The ultimate goals are to better separate concerns, make things easier to test and allow for multiple workspace support.

[Here's the multi-workspace server sample as a reference](https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-multi-server-sample).

### Implementation

The commits are somewhat organized, but it's not a perfect story as I kept hitting more and more issues and had to make changes to fix them. I'll instead explain how the codebase is organized.

- `RubyLsp`: this class represents the entire extension. It owns everything in the extension and controls the top level activation and deactivation. It has a map of workspaces, where each workspace uri points to a `Workspace` instance. Since commands can only be registered once, this class now owns every single command in the extension
- `Workspace`: this class represents a single workspace. It owns the LSP client and the Ruby environment, since each workspace may use different Ruby versions and different dependencies. It's this class' responsibility to activate the Ruby environment, the language server, install the server gem and register restart watchers
- `Ruby`: not many changes here. Each instance of this class represents a Ruby environment activation for a given workspace
- `TestController`: the extension should only have a single test controller, so this class is owned by `RubyLsp`. Its only dependencies are figuring out what workspace it should be running the tests on (because it needs the current working directory and the Ruby environment). We inject a function from `RubyLsp` to allow the controller to select the current workspace
- `Debug`: similar to `TestController`, the extension should only register one debugger client. It uses the same technique of injecting the function to select the current workspace, so that we know the current working directory and the Ruby environment
- `Client`: the client has been refactored to inherit from `LanguageClient`, rather than using the previous decorator pattern by keeping a `client` property. This allows us to have direct access to the client's properties, like state (which is why our own implementation of state was removed)
- `StatusItems`: the recommendation of the VS Code docs is that the language status items should always display the relevant information for the current file, since the items are dependent on which file is opened. Therefore, `RubyLsp` owns the items and refreshes them using an event emitter any time the workspace changes. That allows us to display the workspace-specific information
- `Telemetry`: the only change in telemetry is that it no longer owns the server version. We need that in the `Client`, since each client could be running a different server version

### Automated Tests

This PR is already absolutely immense, so I will follow up with other PRs to add better tests.

#### Single workspace

2. Open a single repository
3. Make sure the following are working
    - Ruby LSP features like diagnostics, formatting, document symbol, etc
    - Start, restart, stop commands
    - Show syntax tree command
    - Run, run in terminal and debug code lens
    - Status items showing the right information. All status items' commands working

#### Multi workspace

2. Open a single repository
3. Click `File` -> `Add folder to workspace`
4. Add a different folder to the current workspace. I recommend using `ruby-lsp` and `tapioca`, so that you can see the different Ruby versions in the status items
5. Make sure the following are working for both workspaces
    - Ruby LSP features like diagnostics, formatting, document symbol, etc
    - Start, restart, stop commands
    - Show syntax tree command
    - Run, run in terminal and debug code lens
    - Status items showing the right information. All status items' commands working

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,7 +1,4 @@
 import path from "path";
-import fs from "fs";
-import { promisify } from "util";
-import { exec } from "child_process";
 import { performance as Perf } from "perf_hooks";
 
 import * as vscode from "vscode";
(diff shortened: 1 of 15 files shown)


st0012:
I tested this against projects under the `sentry-ruby` [repo](https://github.com/getsentry/sentry-ruby):

- Go-to-definition worked well
- CodeLens displayed correctly
    - `Run` and `Debug` had no issues
    - But `Run In Terminal` would reuse the same terminal, which would be created under one workspace's folder, and make the other workspace's test command to fail. Ideally each workspace should have its own dedicated test terminal
- DocumentSymbol worked well
- All traces are still printed to the same output tab. We should maybe prefix the workspace/project info to make it easier to distinguish the source

vinistock:
@st0012 started creating a new terminal for each workspace. I agree with the output feedback, but unfortunately VS Code's output channel has no support for tags of any kind.

That means we can't simply create a separate tag for each workspace, we'd need to manually prefix all logs with the workspace name, which is a bit messy. I think we can try to propose something to VS Code itself.

I also pushed a commit to fix https://github.com/Shopify/vscode-ruby-lsp/issues/381#issuecomment-1804238115. We were activating every workspace regardless of whether they are Ruby projects. We need to try to find lockfiles and also allow for lazy activation for workspaces where there aren't any.

src/status.ts
```diff
@@ -307,89 +183,48 @@ export class FeaturesStatus extends StatusItem {
...
+      if (workspace) {
+        this.items.forEach((item) => item.refresh(workspace));
+      }
+    });
   }
 
-  public refresh() {
-    this.items.forEach((item) => item.refresh());
+  dispose() {
+    this.items.forEach((item) => item.dispose());
```
st0012, line 228: When didn't we call `dispose` before but now need to 🤔

src/rubyLsp.ts
```diff
@@ -0,0 +1,401 @@
...
+    this.telemetry = new Telemetry(context);
+    this.testController = new TestController(
+      context,
+      this.telemetry,
+      this.currentActiveWorkspace.bind(this),
+    );
+    this.debug = new Debugger(context, this.currentActiveWorkspace.bind(this));
+    this.registerCommands(context);
+
+    this.statusItems = new StatusItems();
```
st0012, line 36: It looks like we'll call `extension.activate()` right after this constructor is executed. Is it possible to initialise `StatusItems` there with workspace instead?

src/rubyLsp.ts
```diff
@@ -0,0 +1,401 @@
...
+import DocumentProvider from "./documentProvider";
+import { Workspace } from "./workspace";
+import { Command, STATUS_EMITTER, pathExists } from "./common";
+import { VersionManager } from "./ruby";
+import { StatusItems } from "./status";
+import { TestController } from "./testController";
+import { Debugger } from "./debugger";
+
+// The RubyLsp class represents an instance of the entire extension. This should only be instantiated once at the
+// activation event. This class controls all of the existing workspaces, telemetry and handles all commands
```
st0012, line 16: Does one RubyLsp instance control multiple workspaces, or it's mapped 1 to 1? I think this can be noted in the comment here too.

src/client.ts
```diff
@@ -14,286 +12,172 @@ import {
...
-
-export default class Client implements ClientInterface {
-  private client: LanguageClient | undefined;
-  private readonly workingFolder: string;
-  private readonly telemetry: Telemetry;
-  private readonly statusItems: StatusItems;
-  private readonly testController: TestController;
-  private readonly customBundleGemfile: string = vscode.workspace
+// Get the executables to start the server based on the user's configuration
+function lspExecutables(
```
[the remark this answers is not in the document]
vinistock: Done!

src/rubyLsp.ts
vinistock: If VS Code is opened with multiple workspaces and no files are currently opened, then the current active workspace is `undefined`. That would require handling the `undefined` scenario in the status items, which is why I ended up with this approach.

Basically, the status items are first instantiated with backup values. The moment a file is opened or switched, the active workspace changes and then we immediately update to show the information relevant to that workspace.

src/rubyLsp.ts
vinistock: It controls all of the workspaces. There's only one `RubyLsp` instance, but many `Workspace` instances. I updated the comment, please let me know if it's clear or if I can better word it.

Issue #828, opened 2023-10-12
Modules::include wrong indentation


akirataguchi115:
### Ruby version

ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]

### Code snippet
```ruby
class Brewery < ApplicationRecord
  include RatingAverage
  has_many :beers, dependent: :destroy
end
```
### Description

Paste the code snippet. Remove line 2. Start writing the same line starting with "include" with indentation already input there like in the snippet.

### Expected output

The indentation holds and the include gets typed correctly there.

### Actual output
Ruby LSP indents the "include" backwards as if the include shouldn't be indented.

According to the Official Ruby API `include` should be indented: https://docs.ruby-lang.org/en/master/Module.html

davidcelis:
Not sure if its the _same_ issue or only related, but I've found some other auto indentation issues when using Shopify's Ruby LSP extension. In addition to the original problem around `include` being auto-indented as soon as you type `in` (probably because it tries to detect it as the pattern matching keyword), I've also noticed inconsistent behavior with `do...end` blocks. For example, in RSpec tests, different things happen with different kinds of blocks:

```ruby
describe "an rspec test group" do
# does not properly increase indentation after entering the newline

let(:variable) do
  # properly increases indentation after entering the newline
```

akirataguchi115:
I'll try to reproduce this in the next 48 hours.

akirataguchi115:
I am able to reproduce the bug. Editing the Code snippet to be more readable as the issue template messed it up.

I can confirm that the wrong auto indent already happens, as stated above, when typing "in".

I can also reproduce the following bug mentioned above:
```ruby
describe "an rspec test group" do
# does not properly increase indentation after entering the newline

let(:variable) do
  # properly increases indentation after entering the newline
```

The bug still happens after setting the recommended configurations and disabling `ruby-rubocop` extension.

akirataguchi115:
Great to know that you can reproduce it. I'm not using Copilot though.

davidcelis:
Will that fix the issue in VSCode? It seems like that's against the now-deprecated default Ruby extension, and everybody now is being told to install Shopify's Ruby LSP extension

andyw8:
The change above corresponds to the built-in Ruby Language Basics extension:

[...]

PR #897, opened 2023-11-15
Fix in/when dedent with language configuration

vinistock:
### Motivation

Fixes part of #828

The issue with the indentation is the Ruby language configuration defined inside VS Code itself. We can expose a configuration to fix it temporarily, but we will upstream this to VS Code itself so that we don't have to bundle one.

### Implementation

Instead of decreasing the indentation on `in|when`, we only decrease on `(in|when)\s` which waits for a single space to do it. That way, things like `include` will never match and not reduce the indendation.


andyw8 (approved):

vinistock:
No reviews on the upstream PR yet, so I will merge this one so that we can improve the experience for now.

Once the upstream PR is merged, we can revert this one and just rely on VS Code to do the right thing.

PR #904, opened 2023-11-20
Remove newlines from Ruby activation script

### LINKED ISSUE #891: Failed to activate none environment: Command failed: ruby -rjson -e ... compile error (SyntaxError)
### Operating System

Windows 10

### Ruby version

3.2.2

### Project has a bundle

- [X] Has bundle

no manager

### Description

I have been running Ruby LSP in VS Code for a short while (after the deprecation of the previous Ruby extension), but when I came back to it this morning upon opening my first .rb file I was greeted with:

> Failed to activate none environment: Command failed: ruby -rjson -e "STDERR.printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump({ env: ENV.to_h, ruby_version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT) }))" -e:1: syntax error, unexpected end-of-input, expecting ')' ...V_ACTIVATE%sRUBY_ENV_ACTIVATE}, ... ^ -e: compile error (SyntaxError)

I've run through all the steps on https://github.com/Shopify/vscode-ruby-lsp#troubleshooting without success. 

I also noted https://github.com/Shopify/vscode-ruby-lsp/issues/883, but my Path does not contain any spaces. It is simply `C:\Ruby32-x64\bin`

Running `ruby -rjson -e "STDERR.printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump({ env: ENV.to_h, ruby_version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT) }))"` in a terminal returns the JSON block it appears to be after quite happily, including the appropriate `"ruby_version":"3.2.2"`

The rest of my system knows where Ruby is, what version it is, installs and runs gems, etc. I am fundamentally stumped on what the issue is or where it might have come from.

I have the latest version of the ruby-lsp gem installed, 0.21.3, and the above is the same on the VSCode Release 0.4.17 and Pre-Release 0.5.0 versions.

I also attempted to define a custom activation as suggested in that ticket anyway, following the instructions at https://github.com/Shopify/vscode-ruby-lsp/blob/main/VERSION_MANAGERS.md#custom-activation, but that just led to a different error:

> Failed to activate custom environment: Command failed: PATH=C:\Ruby32-x64\bin:$PATH && ruby -rjson -e "STDERR.printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, JSON.dump({ env: ENV.to_h, ruby_version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT) }))" 'ruby' is not recognized as an internal or external command, operable program or batch file.

I tried every variation of "C:\Ruby32-x64\bin" I could think of (double slashed, reverse slashed, trailing slashes, relative paths, etc) all to the same result.

Happy to provide any additional info or try other potential troubleshooting steps.

vinistock:
### Motivation

Closes #891, closes https://github.com/Shopify/ruby-lsp/issues/1175

Our Ruby activation script was too long, so we broke it into multiple lines. Based on the information shared in the issue, Windows does not seem to accept line breaks in terminal input and that ends up causing a syntax error.

### Implementation

Instead of using a multiline string interpolation, I just changed to write the script in parts inside an array, which we join at the end.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -169,8 +169,16 @@ export class Ruby {
 
   private async activate(ruby: string) {
     let command = this.shell ? `${this.shell} -ic '` : "";
-    command += `${ruby} -rjson -e "STDERR.printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE},
-      JSON.dump({ env: ENV.to_h, ruby_version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT) }))"`;
+
+    // The Ruby activation script is intentionally written as an array that gets joined into a one liner because some
+    // terminals cannot handle line breaks. Do not switch this to a multiline string or that will break activation for
+    // those terminals
+    const script = [
+      "STDERR.printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, ",
+      "JSON.dump({ env: ENV.to_h, ruby_version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT) }))",
+    ].join("");
+
+    command += `${ruby} -rjson -e "${script}"`;
 
     if (this.shell) {
       command += "'";


src/ruby.ts
```diff
@@ -169,8 +169,13 @@ export class Ruby {
 
   private async activate(ruby: string) {
     let command = this.shell ? `${this.shell} -ic '` : "";
-    command += `${ruby} -rjson -e "STDERR.printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE},
-      JSON.dump({ env: ENV.to_h, ruby_version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT) }))"`;
+
+    const script = [
+      "STDERR.printf(%{RUBY_ENV_ACTIVATE%sRUBY_ENV_ACTIVATE}, ",
+      "JSON.dump({ env: ENV.to_h, ruby_version: RUBY_VERSION, yjit: defined?(RubyVM::YJIT) }))",
+    ].join("");
```
andyw8, line 176: May be worth adding a comment here explaining the approach, in case it gets refactored in future and someone is tempted to re-write it as a single string.

andyw8 (approved):

PR #907, opened 2023-11-20
Read server information from initialize result

vinistock:
### Motivation

Running Ruby commands in separate shells to determine server version and formatter is actually an incorrect approach. The [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult) allows the server to return its version during the initialize result.

Additionally, although not listed in the specification docs, servers can actually return [any key value pairs](https://github.com/microsoft/vscode-languageserver-node/blob/02806427ce7251ec8fa2ff068febd9a9e59dbd2f/protocol/src/common/protocol.ts#L1501) as part of the initialize result.

Switching to getting this information from the server reduces duplication and removes the need to shell out to Ruby multiple times.

### Implementation

Started reading server version and formatter from the initialize result.

### DIFF
--- a/src/status.ts
+++ b/src/status.ts
@@ -133,7 +133,9 @@ export class ServerStatus extends StatusItem {
   refresh(): void {
     switch (this.client.state) {
       case ServerState.Running: {
-        this.item.text = `Ruby LSP v${this.client.serverVersion}: Running`;
+        this.item.text = this.client.serverVersion
+          ? `Ruby LSP v${this.client.serverVersion}: Running`
+          : "Ruby LSP: Running";
         this.item.command!.arguments = [STARTED_SERVER_OPTIONS];
         this.item.severity = vscode.LanguageStatusSeverity.Information;
         break;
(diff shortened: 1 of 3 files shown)


andyw8 (approved):
Nice! Will be great to avoid the need to keep all these in sync with the server.

vinistock:
Only the latest version returns it, but I would really like to remove this code as it deviates from the specification and makes us waste time with two shell invocations to Ruby processes.

The only downside for people on older server versions is that the formatter and server version status items will disappear.

We can wait a couple of weeks to give people the chance to upgrade. Would that be okay?

vinistock:
As agreed on Slack, I changed the status items so that they tell people to upgrade if the formatter information is missing.

src/status.ts
```diff
@@ -359,7 +361,11 @@ export class FormatterStatus extends StatusItem {
   }
 
   refresh(): void {
-    this.item.text = `Using formatter: ${this.client.formatter}`;
+    if (this.client.formatter) {
+      this.item.text = `Formatter: ${this.client.formatter}`;
+    } else {
+      this.item.text = `Formatter: upgrade server to v0.12.4 or higher`;
```
st0012, line 367: I worry this could be interpreted as formatter won't work with versions before `v0.12.4`. How about:

```suggestion
      this.item.text = `Formatter: requires server to be v0.12.4 or higher to display this field`;
```

st0012 (approved):

PR #914, opened 2023-11-24
Build test items in hierarchy if it's supported

### LINKED ISSUE #792: Support nested code lens groups
### This feature is VS Code specific

- [X] VS Code specific

### Use case

It will allow [RSpec's code lens](https://github.com/st0012/ruby-lsp-rspec) display correctly in test Explorer. 

### Description

Currently, the `testController` only [supports 1 level of nesting](https://github.com/Shopify/vscode-ruby-lsp/blob/main/src/testController.ts#L127-L134), which generally works fine in minitest/test-unit projects. 

However, RSpec tests are often nested with 2+ level of `context` blocks, which the extension currently can't display well.

### Example

```rb
RSpec.describe Foo do
  it "doest foo" do
  end
  context "when something" do
    it "does bar" do
    end

    context "when another thing" do
      it "does baz" do
      end
    end
  end
end
```

Screenshot 2023-09-17 at 19 22 16

AFAIK in the specification codeLens objects don't have the concept of hierarchy. So we likely will need to implement this by utilising the data attribute with a custom specification.

### Implementation

_No response_

st0012:
### Motivation

Currently, tests in Test Explorer are assumed to only have 2 levels, which is incorrect, especially with RSpec tests.

With a 2 level nesting test file:

```rb
class ParentTest < Minitest::Test
  def test_foo; end

  class FirstChildTest < Minitest::Test
    def test_bar; end
  end

  class SecondChildTest < Minitest::Test
    def test_baz; end
  end

  def test_foobarbaz; end
end
```

The current implementation doesn't display the nesting nor group tests correctly:

Screenshot 2023-11-24 at 20 40 44

But with https://github.com/Shopify/ruby-lsp/pull/1214 and this PR, we can deliever the correct result:

Screenshot 2023-11-24 at 20 42 40

Closes #792

### Implementation

- If code lens items has `group_id` data attribute, we can build items hierarchy with them.
- Otherwise, fallback to the current approach.

### DIFF
--- a/src/testController.ts
+++ b/src/testController.ts
@@ -93,7 +93,9 @@ export class TestController {
       this.testCommands.delete(test);
     });
 
+    const groupIdMap: { [key: string]: vscode.TestItem } = {};
     let classTest: vscode.TestItem;
+
     const uri = vscode.Uri.from({
       scheme: "file",
       path: response[0].command!.arguments![0],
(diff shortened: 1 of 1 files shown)


src/testController.ts
```diff
@@ -121,16 +123,40 @@ export class TestController {
...
+          testItem.tags = [...testItem.tags, this.debugTag];
+        }
+        // If the server doesn't support code lens hierarchy, all groups are top-level
       } else {
-        classTest = testItem;
-        classTest.canResolveChildren = true;
-        this.testController.items.add(testItem);
+        // Add test methods as children to the test class so it appears nested in Test explorer
+        // and running the test class will run all of the test methods
+        // eslint-disable-next-line no-lonely-if
```
st0012, line 151: I feel blending the 2 versions of implementation together makes things harder to understand and also harder to remove the fallback. So I want to separate them with simple if/else condition and simply remove the fallback when we're ready to do so.

vinistock:
Very clean, the implementation looks good to me. Do you have a project that we can manually test this on? Does IRB use nested test declarations?

st0012:
The only minitest/test-unit file that I have to test against is [this](https://github.com/ruby/reline/blob/master/test/reline/test_face.rb).

### Before (incorrect nesting)

Screenshot 2023-11-27 at 17 36 28

### After

Screenshot 2023-11-27 at 17 53 46

vinistock (approved):

PR #918, opened 2023-11-27
Write the Ruby environment information into a temporary file

vinistock:
### Motivation

Closes Shopify/ruby-lsp#1712

Some shell plugins take the stderr pipe over completely and prevent the NodeJS process from properly writing to the stderr in the shell.

One example reported by users was https://github.com/marlonrichert/zsh-snap/blob/7741929695f0e4761971e951ab81cb7c2bd9910f/functions/.znap.prompt#L29. The shell plugin redirects stderr into a temporary file and only brings back stderr if it detects that the user is executing a command. I'm pretty sure the issue is that it cannot detect this when the command is coming from a NodeJS process.

We need to avoid being susceptible to issues like this since we have no control over what the shell or plugins do with the pipes.

This is only a problem for running the Ruby activation command with the interactive shell since it loads files like `~/.zshrc` or `~/.bashrc`, which in turn load plugins that can take over the pipes. We don't have to worry about this in exec calls that do not use the shell interactively.

### Implementation

Started writing the activation results into a temporary file instead of printing to stderr.

### Manual tests

Start the extension in a project that's using `chruby`. Verify that it gets activated properly.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -1,5 +1,6 @@
 import path from "path";
 import fs from "fs/promises";
+import os from "os";
 
 import * as vscode from "vscode";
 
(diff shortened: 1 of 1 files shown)


st0012 (approved):
Confirmed it's working with chruby in Reline and IRB

PR #928, opened 2023-12-04
Update README under custom gemfile section

luisAzcuaga:
### Motivation

Closes https://github.com/Shopify/vscode-ruby-lsp/issues/921

### Implementation

I added an example of how to solve a common problem when using custom gem file.

### Automated Tests

Doesn't apply :)

### Manual Tests

My markdown changes didn't break the readme :)
Here is the evidence.
image


README.md
```diff
@@ -161,6 +161,18 @@ gem "ruby-lsp"
 gem "rubocop"
 ```
 
+> [!NOTE]
+>
+> Take in mind that formatters, linters and their extensions should be included in the custom gem file; you might need to add more gems than the ones shown above.
```
vinistock, line 166:
```suggestion
> Take in mind that formatters, linters and their extensions should be included in the custom gemfile; you might need to add more gems than the ones shown above.
```

README.md
```diff
@@ -161,6 +161,18 @@ gem "ruby-lsp"
...
+> e.g: If you are using rubocop, you would also need to add these gems:
```
vinistock, line 167:
```suggestion
> e.g: If you are using rubocop extensions, you would also need to add them:
```

vinistock (commented):
Two minor comments, but this is great. Thank you!

luisAzcuaga:
> Two minor comments, but this is great. Thank you!

Perfect, I just pushed the changes! And got my last commit verified (that's why I had too many forced pushes before 😅 ) do you want me to squash so it becomes one single and verified commit?

PR #950, opened 2023-12-18
Improve debug workspace selection

vinistock:
### Motivation

There are some edge cases for selecting the active workspace while launching the debugger that we didn't account for before.

For example, if the user has the output tab selected, because VS Code treats that panel as an editor, we receive it as the currently active editor. However, if we try to find a workspace with that URI, we find nothing because the output tab doesn't belong to any workspace.

We should instead rely on the debugger client mechanism better. We receive the current workspace folder as a parameter and we can inject that into the configuration to avoid having to detect the active workspace.

### Implementation

Started relying on the workspace folder informed by VS Code rather than trying to detect the workspace ourselves.

### Manual Tests

Ensure launch configurations and the debug code lens are still working. Then try this:

1. Start the extension on this branch and with the `ruby-lsp` project
2. Select the output tab
3. Try to run the attach launch configuration
4. Verify that you properly get prompted with the list of sockets

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -14,13 +14,15 @@ export class Debugger
 {
   private debugProcess?: ChildProcessWithoutNullStreams;
   private readonly console = vscode.debug.activeDebugConsole;
-  private readonly currentActiveWorkspace: () => Workspace | undefined;
+  private readonly workspaceResolver: (
+    uri: vscode.Uri,
+  ) => Workspace | undefined;
 
   constructor(
     context: vscode.ExtensionContext,
-    currentActiveWorkspace: () => Workspace | undefined,
+    workspaceResolver: (uri: vscode.Uri) => Workspace | undefined,
   ) {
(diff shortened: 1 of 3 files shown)


src/debugger.ts
```diff
@@ -14,13 +14,13 @@ export class Debugger
 {
   private debugProcess?: ChildProcessWithoutNullStreams;
   private readonly console = vscode.debug.activeDebugConsole;
-  private readonly currentActiveWorkspace: () => Workspace | undefined;
+  private readonly getWorkspace: (uri: vscode.Uri) => Workspace | undefined;
 
   constructor(
     context: vscode.ExtensionContext,
-    currentActiveWorkspace: () => Workspace | undefined,
+    getWorkspace: (uri: vscode.Uri) => Workspace | undefined,
```
andyw8, line 21: (nit) The `get` prefix may suggest this is an accessor for `workspace`... could this instead be something like `resolveWorkspace`?

andyw8 (approved):

PR #961, opened 2024-01-04
Fix Run Current Test In Terminal for class line

snutij:
### Motivation

Closes: #799 

### Implementation

Previously, given this code 

```ruby
class Foo # cursor here
 
  def test_one
    assert(true)
  end

end
```

The test item called 'Foo' has children called 'test_one'. We set the 'testItem' to run with the value of the recursive method `findTestByActiveLine`. But in this case this will be `undefined`, as the cursor is not in range of `test_one`. And as this is an `else_if` condition, we do not go through the assignation latter.

I just reworked condition: 
- We assign the `testItem` with this `Foo` test. 
- If there is a child, we call the method again with children subset
  - if a child match the range, we assign the value (between `Foo` and `test_one`, both can match the cursor range, but child have priority)
  - if no child match the range, we do not assign `undefined`, like that we keep the first assignation 
  
### Automated Tests

There are no automated tests for TestController, do you think it's worth it to avoid this kind of regression? 

### Manual Tests

Checkout this branch, and run tests with command with cursor at the top level of group (on class level for example)

<details><summary>record</summary>

Check the command output, first, `test_one` level, is suffixed by `:7`, others are the top level command.

Jan-04-2024 20-31-21

</details>

### DIFF
--- a/src/testController.ts
+++ b/src/testController.ts
@@ -400,14 +400,19 @@ export class TestController {
     testItems.forEach((test) => {
       if (testItem) return;
 
-      if (test.children.size > 0) {
-        testItem = this.findTestByActiveLine(editor, test.children);
-      } else if (
+      if (
         test.uri?.toString() === editor.document.uri.toString() &&
         test.range?.start.line! <= line &&
         test.range?.end.line! >= line
       ) {
         testItem = test;
+      }
+
+      if (test.children.size > 0) {
+        const childInRange = this.findTestByActiveLine(editor, test.children);
+        if (childInRange) {
+          testItem = childInRange;
+        }
       }
     });
 


src/testController.ts
```diff
@@ -400,15 +400,20 @@ export class TestController {
...
-      } else if (
+      if (
         test.uri?.toString() === editor.document.uri.toString() &&
         test.range?.start.line! <= line &&
         test.range?.end.line! >= line
       ) {
         testItem = test;
       }
+
+      if (test.children.size > 0) {
```
KaanOzkan, line 411: Couldn't this be an `else if`? Am I misunderstanding the conditional on line 403?

snutij: No, this is not possible, for example: 

```ruby
class Foo 
 
  def test_one # cursor here
    assert(true)
  end

end
```

The first condition will match (parent test range match the cursor) but you want to override the `testItem` value with child. Is it clearer?

Issue #537, opened 2023-04-04
Unable to disable semantic highlighting


ryanb:
I have `semanticHighlighting` config set to `false` yet it's still overriding the default syntax highlighting grammars. I believe this is due to the grammar file built into the extension from #187.

Is there a way to disable this?

vinistock:
The problem is that Ruby's syntax is ambiguous and we can never correctly highlight it without the server. For example, if you have a line that just has `foo` in it, it's impossible to tell if it's a local variable or method invocation with a regex (which is what grammars use).

That's where semantic highlighting comes in, it parses the Ruby file and determines what exactly that piece of code is supposed to be. Is there any reason why you can't use semantic highlighting?

In my experience, most themes don't do a great job at highlighting Ruby. If I may recommend a different one, our [extension pack](https://github.com/Shopify/vscode-shopify-ruby#themes) bundles a theme that uses all of the Ruby LSP semantic highlighting information.

ryanb:
@vinistock

> That's where semantic highlighting comes in, it parses the Ruby file and determines what exactly that piece of code is supposed to be. Is there any reason why you can't use semantic highlighting?

I plan to. I just need to configure my theme to support it. I'll check out the extension pack, thanks for the link!

As for this issue, it looks like the [motivation for adding a grammar file](https://github.com/Shopify/vscode-ruby-lsp/issues/139) is to provide a fallback for when the LSP isn't available. Can this be done with a separate extension? This way one can opt-in to it or use an alternative extension for grammars if they prefer.

If that isn't an option, are you open to improvements to the grammar file?

PR #962, opened 2024-01-05
Remove experimental variable guessing in Ruby grammar

ryanb:
### Motivation

The grammar file incorrectly assumes most top level method calls are variables.

230160734-f81fc579-5ff1-4760-b0db-3dc5dda01ee7

There is a comment in the grammar labeling this as "experimental".

> This is kindof experimental. There really is no way to perfectly match all regular variables, but you can pretty well assume that any normal word in certain curcumstances that haven't already been scoped as something else are probably variables, and the advantages beat the potential errors

I don't agree with "any normal word in certain curcumstances that haven't already been scoped as something else are probably variables" because method calls fall under this. I'd rather we error on the side of not scoping something we don't know.

When the LSP is active it adds semantic highlighting tokens which properly discerns local variables from method calls. It is better to use this instead of the grammar to determine local variables.

See further discussion in #537.

### Implementation

I removed this experimental part of the grammar so it no longer scopes method calls as `variable.other.ruby`.

### Automated Tests

There are no automated tests for the grammar file.

### Manual Tests

You can check this by opening a Ruby file, putting text cursor in a method call, and running the "Inspect Editor Tokens and Scopes" command. It should not show `variable.other.ruby` under `textmate scopes` section.

### DIFF
--- a/grammars/ruby.cson.json
+++ b/grammars/ruby.cson.json
@@ -2313,11 +2313,6 @@
           "include": "$self"
         }
       ]
-    },
-    {
-      "comment": "This is kindof experimental. There really is no way to perfectly match all regular variables, but you can pretty well assume that any normal word in certain curcumstances that haven't already been scoped as something else are probably variables, and the advantages beat the potential errors",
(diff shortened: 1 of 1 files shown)


vinistock (approved):
Agreed. We haven't revisited this grammar, but I the ideal state would be to not push any scopes for ambiguous syntax

PR #968, opened 2024-01-08
Improve non-ruby workspace folder UX

parsley42:
This PR updates the "no lock file" warning UX to what I would expect given the original comments. Previously "Don't show again" still resulted in starting a ruby-lsp server for the folder even if there was no lock file. Now if there's no lockfile:
* If disableMultirootLockfileWarning is true, it just skips starting the workspace, otherwise it gives the warning
* Yes sends the user to the docs and doesn't start the workspace
* Don't show again doesn't start the workspacea and updates disableMultirootLockfileWarning
* No allows the workspace to start even without a lockfile

### Motivation

My standard workspace includes a ruby and non-ruby root. Prior to this PR, there was no way to tell the extension not to start ruby-lsp for the non-ruby root and not to ask again.

### Implementation

I tried to match the behavior of activateWorkspace to the original comment:
`If one of the workspaces does not contain a lockfile, then we don't try to start a language server.` Especially, clicking "Don't ask again" doesn't start the ruby-lsp server for the directory and doesn't give a warning.

### Automated Tests

I honesn't don't know how to write the tests - I'm a devops engineer, not a javascript programmer, and it took me a long time to write this as-is.

### Manual Tests

I built and installed my version of the plugin, and now all the possible responses to the "no lockfile" warning behave as expected - importantly, I say "don't ask again" and the ruby-lsp server won't start for the given directory.

### Final Thoughts

I kind of expect I'm missing something here, and am preparing myself for a major "D'oh!" moment.

### DIFF
--- a/src/rubyLsp.ts
+++ b/src/rubyLsp.ts
@@ -117,18 +117,21 @@ export class RubyLsp {
     // If one of the workspaces does not contain a lockfile, then we don't try to start a language server. If the user
     // ends up opening a Ruby file inside that workspace, then we lazily activate the workspace. These need to match our
     // `workspaceContains` activation events in package.json
-    if (
-      customBundleGemfile.length === 0 &&
-      !(await pathExists(path.join(workspaceDir, "Gemfile.lock"))) &&
-      !(await pathExists(path.join(workspaceDir, "gems.locked"))) &&
-      !this.context.globalState.get("rubyLsp.disableMultirootLockfileWarning")
-    ) {
+    var startWorkspace = customBundleGemfile.length !== 0 ||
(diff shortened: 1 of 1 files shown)


vinistock:
Thank you for the contribution and the feedback on the experience. You're right, it's odd right now.

I was trying to create a picture in my head of the different scenarios and ended up creating a PR to differentiate eager and lazy activation, which I believe fixes this issue https://github.com/Shopify/vscode-ruby-lsp/pull/971.

Can you let me know what you think about it? Does it cover the scenario you brought up?

parsley42:
> Can you let me know what you think about it? Does it cover the scenario you brought up?

Yup, your PR is definitely UX++ over this one. I'll close, thanks.

PR #976, opened 2024-01-11
Fallback to guessing the active workspace if none is found based on URI

### LINKED ISSUE #947: Debugging requires a workspace folder to be opened
### Operating System

Mac + Remote linux server over SSH 

### Ruby version

2.7.8

### Project has a bundle

- [ ] Has bundle

no manager

### Description

I am running on a remote linux machine over SSH using a code workspace file. 

The file is in my user home directory with the following contents.

```
  "folders": [
    {
      "path": "/home/myuser"
    }
  ]
```

My debug configuration is as follows:

```
        {
          "name": "Debug test",
          "type": "ruby_lsp",
          "request": "launch",
          "program": "ruby /home/myuser/.../test.rb",
        },
```

The script open in my editor. When clicking debug, I receive an error message `Debugging requires a workspace folder to be opened` no matter what. The folder is open in my workspace. Even if I set the folder to open the exact folder where test.rb is placed, I still receive the same message. I'm not sure if this is caused because this is a remote development environment or something wrong with my configuration.

vinistock:
### Motivation

Closes #947

I believe I understand the problem some users are having. In #950, we changed the logic to not try to guess the active workspace anymore and just rely on the folder passed by VS Code itself.

However, I think this doesn't work for monorepos that place their `launch.json` configurations on the top level. This project structure:

```
monorepo/
  .vscode/launch.json
  client/
  server/
```

For this scenario, even with the multi-root workspace configuration, VS Code gives us the URI for the top level when launching the debugger because that's where the debugging task is configured. But since that's not a Ruby workspace, we never find anything.

If we fallback to guessing the active workspace, we should be able to account for this scenario better.

### Implementation

Created a new method to resolve the workspace that accepts a nilable URI. The key difference in behaviour is that if the URI is present, but we can't find a workspace for it, we now fallback to trying to figure out the active workspace (which we weren't doing).

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -15,12 +15,12 @@ export class Debugger
   private debugProcess?: ChildProcessWithoutNullStreams;
   private readonly console = vscode.debug.activeDebugConsole;
   private readonly workspaceResolver: (
-    uri: vscode.Uri,
+    uri: vscode.Uri | undefined,
   ) => Workspace | undefined;
 
   constructor(
     context: vscode.ExtensionContext,
-    workspaceResolver: (uri: vscode.Uri) => Workspace | undefined,
+    workspaceResolver: (uri: vscode.Uri | undefined) => Workspace | undefined,
   ) {
     this.workspaceResolver = workspaceResolver;
 
@@ -86,14 +86,12 @@ export class Debugger
     debugConfiguration: vscode.DebugConfiguration,
     _token?: vscode.CancellationToken,
   ): vscode.ProviderResult<vscode.DebugConfiguration> {
-    if (!folder) {
-      throw new Error("Debugging requires a workspace folder to be opened");
-    }
-
-    const workspace = this.workspaceResolver(folder.uri);
+    const workspace = this.workspaceResolver(folder?.uri);
 
     if (!workspace) {
-      throw new Error(`Couldn't find workspace ${folder.name}`);
+      throw new Error(
+        `Couldn't find a workspace for URI: ${folder?.uri} or editor: ${vscode.window.activeTextEditor}`,
+      );
     }
 
     if (debugConfiguration.env) {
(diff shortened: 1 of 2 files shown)


src/rubyLsp.ts
```diff
@@ -359,6 +359,28 @@ export class RubyLsp {
...
+    if (workspace) {
+      return workspace;
+    }
+
+    // Otherwise, if there's a URI, but we can't find a workspace for it, we fallback to trying to figure out what the
+    // active workspace is. This situation may happen if we receive a workspace folder URI that is not the actual
+    // workspace where the Ruby application exists. For example, if you have a monorepo with client and server
+    // directories and the `launch.json` file is in the top level directory, then we may receive the URI for the top
+    // level, but the actual workspace is the server directory
+    return this.currentActiveWorkspace();
```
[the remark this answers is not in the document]
st0012: I see, thanks for clarification 👍

egiurleo (approved):

src/debugger.ts
```diff
@@ -165,13 +163,14 @@ export class Debugger
...
     let initialized = false;
 
-    const workspaceFolder = session.workspaceFolder;
-    if (!workspaceFolder) {
-      throw new Error("Debugging requires a workspace folder to be opened");
+    const cwd = session.workspaceFolder?.uri.fsPath;
+    let workspaceName = session.workspaceFolder?.name;
+
+    if (!workspaceName) {
+      workspaceName = cwd ? path.basename(cwd) : "unknown";
```
st0012, line 170: Should we error out when the workspace name is unknown? Is it possible to enter this case but still can use the debugger correctly 🤔

vinistock: That's a good point. I started saving our resolved workspace in the debug session configuration instead, so that we can reuse it when spawning the debugger.

That way, the error handling happens when resolving debug configurations and spawning just uses the information saved directly.

PR #979, opened 2024-01-12
Add workspace logger

vinistock:
### Motivation

Closes https://github.com/Shopify/ruby-lsp/issues/1298

When we added multi-root workspace support, we discussed that it would be nice to have each language server be tagged with their workspace name in the logs. Back then I investigated it, but the VS Code API doesn't include anything we can use for that out of the box.

When that issue was created, it got me thinking again and I figured we could use the decorator pattern to wrap the actual logger and prefix all operations with the workspace name.

Ideally, VS Code's logger would support tags or something of the sort, but I think this is decent enough while that doesn't happen.

### Implementation

Basically, there's a new `WorkspaceChannel` class that wraps the real logger and stores the workspace name. Then all operations are prefixed with `(WORKSPACE_NAME)`, so that it's easier to tell where things are coming from.

Everything else is delegated to the logger directly.

### Automated Tests

Added a test to verify the prefix is properly inserted.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -14,9 +14,10 @@ import {
   MessageSignature,
 } from "vscode-languageclient/node";
 
-import { LOG_CHANNEL, LSP_NAME, ClientInterface } from "./common";
+import { LSP_NAME, ClientInterface } from "./common";
 import { Telemetry, RequestEvent } from "./telemetry";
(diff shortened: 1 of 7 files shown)


egiurleo (approved):
Nice use of the decorator pattern! This is cool.

PR #980, opened 2024-01-12
Invoke shell command with separated arguments

GregCornis:
Cf Shopify/vscode-ruby-lsp#978

To support nushell, which doesn't accept having both arguments at the same time. Fish, ZSH, Bash and sh are all compatible with this different syntax

### Manual Tests

⚠️ I haven't tested it locally. However I made sure that all common shells are compatible : Zsh, Fish, Sh and Bash.

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -168,7 +168,7 @@ export class Ruby implements RubyInterface {
   }
 
   private async activate(ruby: string) {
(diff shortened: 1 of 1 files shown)


vinistock:
We would need to do the same thing in the `toolExists` method, which also invokes the shell.

vinistock (approved):
Thank you for the contribution! Tested it on bash, zsh and fish and everything seems fine

PR #983, opened 2024-01-15
Move Dev Container hint from troubleshooting to readme

Earlopain:
### Motivation

Dev Containers are great but I wasn't previously aware of them (https://github.com/Shopify/ruby-lsp/issues/1303). Documentation for this already exists but not in the place I would expect it to.

### Implementation
Move the text from troubleshooting to readme.
I've also added a note that only Docker is officially supported by the Dev Container extension.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -100,16 +100,7 @@ steps that led to the server getting stuck.
 
 ### Developing on containers
 
-The Ruby LSP is a detached language server, which means it's a background process that runs separately from the VS Code
-instance. To provide its functionality, the Ruby LSP must be running in the same place where your project files exist
-and dependencies are installed.
-
-VS Code supports connecting to containers out of the box, which makes all editor features work seamlessly. That includes
-language servers, the integrated terminal, etc. The VS Code documentation has instructions on how to develop on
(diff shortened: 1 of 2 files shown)


TROUBLESHOOTING.md
```diff
@@ -98,19 +98,6 @@ that prevents it from responding to new requests coming from the editor. If you
 report [here](https://github.com/Shopify/vscode-ruby-lsp/issues/new?labels=bug&template=bug_template.yml) including the
 steps that led to the server getting stuck.
 
-### Developing on containers
```
vinistock, line 101: Let's keep a link to the docs in troubleshooting because we link directly to this document from an error dialog.

Something like

```
### Developing on containers

See the [documentation]().
```

Earlopain: :+1:

connorshea:
@vinistock As part of the ruby-lsp monorepo switch, I think this bit of documentation was lost? [The link that's supposed to point to container docs in TROUBLESHOOTING.md is busted now](https://github.com/Shopify/ruby-lsp/blob/main/TROUBLESHOOTING.md#developing-on-containers).

Could we get that fixed? :)

PR #989, opened 2024-01-17
Add debugger launch test

vinistock:
### Motivation

We had a few debugger related issues recently, so we need to continue on our quest to improve our test coverage of the extension. This PR adds a launch test for the debugger that actually spawns it, so that we can catch more issues on CI.

### Implementation

The test performs all of the necessary setup, like creating a Gemfile and activating Ruby. The it attempts to launch the debugger with a simple `1 + 1` Ruby program, waits for the debugger to finish and performs clean up.

I also bumped our Ruby version to 3.3.0.

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -1,11 +1,25 @@
 import path from "path";
 import fs from "fs";
-import { ChildProcessWithoutNullStreams, spawn, execSync } from "child_process";
+import net from "net";
+import os from "os";
+import { ChildProcessWithoutNullStreams, spawn } from "child_process";
 
 import * as vscode from "vscode";
 
-import { LOG_CHANNEL } from "./common";
+import { LOG_CHANNEL, asyncExec } from "./common";
 import { Workspace } from "./workspace";
+
+class TerminalLogger {
+  append(message: string) {
+    // eslint-disable-next-line no-console
+    console.log(message);
+  }
+
+  appendLine(value: string): void {
+    // eslint-disable-next-line no-console
+    console.log(value);
+  }
+}
 
 export class Debugger
   implements
@@ -13,7 +27,11 @@ export class Debugger
     vscode.DebugConfigurationProvider
 {
   private debugProcess?: ChildProcessWithoutNullStreams;
-  private readonly console = vscode.debug.activeDebugConsole;
+  // eslint-disable-next-line no-process-env
+  private readonly console = process.env.CI
+    ? new TerminalLogger()
+    : vscode.debug.activeDebugConsole;
+
   private readonly workspaceResolver: (
     uri: vscode.Uri | undefined,
   ) => Workspace | undefined;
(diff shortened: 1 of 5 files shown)


src/test/suite/debugger.test.ts
```diff
@@ -134,4 +136,74 @@ suite("Debugger", () => {
...
+  test("Launching the debugger", async () => {
+    // eslint-disable-next-line no-process-env
+    if (process.env.CI) {
+      await vscode.workspace
+        .getConfiguration("rubyLsp")
+        .update("rubyVersionManager", VersionManager.None, true, true);
+    }
+
+    // By default, VS Code always saves all open files when launching a debugging session. This is a problem for tests
+    // because it attempts to save an untitled test file and then we get stuck in the save file dialog with no way of
```
andyw8, line 149: Which file is the untitled one here?

vinistock: It's a new untitled file that gets created when we run tests.

src/test/suite/debugger.test.ts
```diff
@@ -134,4 +136,74 @@ suite("Debugger", () => {
...
+    // the termination callback or else we try to dispose of the debugger client too early
+    vscode.debug.onDidTerminateDebugSession(async (_session) => {
+      debug.dispose();
+      context.subscriptions.forEach((subscription) => subscription.dispose());
+      fs.rmSync(tmpPath, { recursive: true, force: true });
+      await vscode.workspace
+        .getConfiguration("debug")
+        .update("saveBeforeStart", currentSaveBeforeStart, true, true);
+    });
+  }).timeout(30000);
```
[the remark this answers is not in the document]
vinistock: We can definitely do it, but since it's not used anywhere else I'm now sure what to extract. Do you have a suggestion?

src/debugger.ts
```diff
@@ -192,16 +209,19 @@ export class Debugger
     const configuration = session.configuration;
     const workspaceFolder = configuration.targetFolder;
     const cwd = workspaceFolder.path;
+    const port =
```
vinistock, line 212: Using UNIX sockets doesn't work on Windows, but we can easily spawn the debugger with a port.

src/debugger.ts
```diff
@@ -253,7 +277,7 @@ export class Debugger
       // If the Ruby debug exits with an exit code > 1, then an error might've occurred. The reason we don't use only
       // code zero here is because debug actually exits with 1 if the user cancels the debug session, which is not
       // actually an error
-      this.debugProcess.on("exit", (code) => {
+      this.debugProcess.on("close", (code) => {
```
vinistock, line 280: The `exit` event actually happens before the stdin, stdout and stderr pipes are closed. If we raise an error on `exit`, we never capture the backtrace of the server printed to `stderr`.

But if we use the `close` event, then everything gets printed and it's significantly easier to understand why the debug server failed to start.

KaanOzkan (approved):
It might be good to have a timeout for the test if there isn't any reasonable one.

andyw8 (approved):

PR #1002, opened 2024-01-25
Make Ruby version detection fall through each version source

paracycle:
### Motivation

Currently Ruby version detection fails early when it is able to find a particular kind of Ruby version file, but is unable to extract a valid Ruby version from the file. This commit makes it so that Ruby version detection falls through each version source until it finds a valid Ruby version, or fails with a message that no valid Ruby version could be found.

### Automated Tests

Added tests for edge cases with `dev.yml` and fallbacks.

### Manual Tests

Add/remove malformed `.ruby-version` or `dev.yml` files in the presence of a well-formed `.tool-versions` file and observe that the correct version is picked up every single time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -463,6 +463,7 @@
     "@types/glob": "^8.1.0",
     "@types/mocha": "^10.0.6",
     "@types/node": "20.x",
+    "@types/sinon": "^17.0.3",
     "@types/vscode": "^1.68.0",
     "@typescript-eslint/eslint-plugin": "^5.62.0",
     "@typescript-eslint/parser": "^5.62.0",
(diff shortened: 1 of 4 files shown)


vinistock (approved):
Nice.

In addition to this, can we move the `.ruby-version` check to be the first one?

We were preferring `dev.yml`, but since we're moving away from it and `.ruby-version` is the most common one among other version managers, we might as well put it at the top.

paracycle:
> Nice.
> 
> In addition to this, can we move the `.ruby-version` check to be the first one?
> 
> We were preferring `dev.yml`, but since we're moving away from it and `.ruby-version` is the most common one among other version managers, we might as well put it at the top.

Sure, makes sense. Will do!

vinistock (approved):
🚀

PR #1007, opened 2024-01-26
Only eagerly activate the first workspace folder

vinistock:
### Motivation

Closes #975

On setups including many workspace folders, we cause a performance issue because we eagerly activate the Ruby LSP on every one.

We can avoid this by making the eager activation occur only in the first workspace and every other workspace lazy. That way, people using a single workspace start indexing early, maintaining the same experience.

The lazy workspaces get activated as soon as a Ruby document is opened in the VS Code UI.

### Implementation

- Removed the eager activation when switching workspace folders
- Started eagerly activating only the first workspace folder

### DIFF
--- a/src/rubyLsp.ts
+++ b/src/rubyLsp.ts
@@ -53,11 +53,6 @@ export class RubyLsp {
           this.workspaces.delete(workspaceFolder.uri.toString());
         }
       }
-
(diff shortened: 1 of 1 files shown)


andyw8 (approved):
Didn't :tophat: but makes sense.

PR #1011, opened 2024-01-29
Convert test output newlines to \r\n

erikced:
### Motivation

As is documented in the [testing api documentation](https://code.visualstudio.com/api/extension-guides/testing#test-output) VSCode expects test output to have `\r\n` newlines, which is not the default value for many tools and it is up to the extension to convert them. 

### Implementation

N/A

### Automated Tests

Not implemented, `TestController` currently lacks unit tests.

### Manual Tests

Run a ruby unit test with multiline output, compare output with a run executed in the terminal.

### DIFF
--- a/src/testController.ts
+++ b/src/testController.ts
@@ -289,7 +289,7 @@ export class TestController {
             workspace.ruby.env,
           );
 
-          run.appendOutput(output, undefined, test);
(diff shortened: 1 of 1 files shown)


erikced:
I have signed the CLA!

vinistock (approved):

PR #1018, opened 2024-02-01
Add `Dependencies` editor view

paracycle:
### Motivation

This PR adds a DEPENDENCIES view to the Editor sidebar, which shows the gem dependencies (including transitive ones) of a Ruby project and allows users to browse the contents of those dependencies without leaving their current editor.

image

### Implementation

Most of the implementation is centred around the `dependenciesTree.ts` file which declares a `TreeDataProvider` class with nested `TreeItem` types under it. The `DependenciesTree` class watches for changes in workspace, and when a new workspace is loaded, it makes the `rubyLsp/workspace/dependencies` LSP request to the server. Using the response, builds the appropriate dependency tree.

### Automated Tests

No tests yet.

### Manual Tests

Open the current branch against Ruby LSP that includes changes in https://github.com/Shopify/ruby-lsp/pull/1346 and you should be able to see a `DEPENDENCIES` view in the Editor sidebar.

### DIFF
--- a/src/dependenciesTree.ts
+++ b/src/dependenciesTree.ts
@@ -29,7 +29,7 @@ export class DependenciesTree
     });
 
     this.workspaceListener = STATUS_EMITTER.event((workspace) => {
-      if (!workspace) {
+      if (!workspace || workspace === this.currentWorkspace) {
         return;
       }
 


src/rubyLsp.ts
```diff
@@ -32,6 +34,8 @@ export class RubyLsp {
       this.currentActiveWorkspace.bind(this),
     );
     this.debug = new Debugger(context, this.workspaceResolver.bind(this));
+    // eslint-disable-next-line no-new
+    new DependenciesTree(context);
```
paracycle, line 38: We still need to figure out how to properly construct and then dispose of this object.

As things stand right now, we can't assign this to any variable, since we have no use for the variable itself (i.e. we don't do anything with it, it is just a view).

But then ESLint shouts at us if we just create an object and don't hold on to it (which is also reasonable).

I think the correct thing to do is to make `DependenciesTree` disposable, and then dispose of it when `RubyLsp` gets disposed. The problem is `RubyLsp` is itself not disposable yet.

PR #1019, opened 2024-02-02
Add ability to show current file in Dependencies tree view

paracycle:
### Motivation

It would be nice if the Dependencies tree view revealed the current file, similar to how the standard editor tree works.

### Implementation

By watching the `onDidChangeActiveTextEditor` event, we can reveal the current file in the Dependencies tree view. To be able to do that, however, we need to implement `getParent` method on `DependenciesTree` class. This method is used by VS Code to find the parent of a given node. In our case, we want to return the parent of a given file node, which is either a `GemDirectoryNode` or a `Dependency` node, depending on whether the parent directory is the root path of a gem or not.

Moreover, when we can `TreeView.reveal` method to reveal a node, VSCode forces the tree view to be visible, which is a little jarring if the user has hidden the tree view. To avoid that, we store the current item to reveal in an instance variable, and reveal it when the tree view becomes visible again. If the tree view is already visible, we reveal the item immediately.

### Automated Tests
No tests yet.

### Manual Tests
1. Open the editor and wait for the `Dependencies` view to load.
2. Open a file under the `Dependencies` view.
3. Open another file under the `Dependencies` view.
4. Switch to file opened on Step 3. Observe that that file is now focused on the `Dependencies` view.
5. Make the `Dependencies` view invisible by collapsing it or moving it to another pane on its own.
6. Now switch to the editor opened on Step 4. Observe that the `Dependencies` view does not suddenly become visible.
7. Make the `Dependencies` view visible. Observe that the file from Step 4 is now focused.


andyw8 (approved):
🎩 done, all seems good!

PR #1033, opened 2024-02-16
Only show dependencies tree on Ruby projects

vinistock:
### Motivation

I noticed that we were showing the dependencies tree even on non-Ruby projects, which is not what we want.

### Implementation

I fixed it using the when context API. We set a specific context value when we successfully activate a Ruby workspace and that allows the view to be presented. Otherwise, it's always hidden.


src/rubyLsp.ts
```diff
@@ -169,6 +169,14 @@ export class RubyLsp {
 
     await workspace.start();
     this.context.subscriptions.push(workspace);
+
+    // If we successfully activated a workspace, then we can start showing the dependencies tree view. This is necessary
+    // so that we can avoid showing it on non Ruby projects
+    vscode.commands.executeCommand(
+      "setContext",
+      "rubyLsp.showDependenciesTree",
```
paracycle, line 177: Can we not make this specific to the dependencies tree? We might want to turn on and off other features depending on whether we activated a Ruby project or not.

What about something like: `rubyLsp.activated`

vinistock: Good point! This event will definitely be generally useful. I switched to `activated`.

paracycle (approved):

